PR #1, opened 2020-06-23
Add bigger AWG Sizes

Supern00dle:
Adds wire sizes of 35mm² and 50mm² to the AWG conversion.
Values based on german Wikipedia: https://de.wikipedia.org/wiki/American_Wire_Gauge


pandorazboxx:
Why not even bigger sizes? See https://www.multicable.com/resources/reference-data/cross-reference-awg-to-mm2/

Maybe I'm an oddball but I use 350MCM cables.

PR #5, opened 2020-06-24
Feature: hide pins that are not connected

elliotmr:
Very cool project here, exactly what I needed for a project that I am working on. One issue is that I have a 35-pin connector that I am only going to populate 5-10 pins from. I don't really want to print out 30 unconnected pins on the schematic.

This pull request adds a connector option "hide_disconnected_pins" to hide any pins on the connector that don't have a connection on them. I also added an example as well to see how it works.


formatc1702:
This is super useful, I'll merge it in the next few days. Thanks!

formatc1702:
I've merged it into its own branch, will need to adapt it to make it work with the changes from #10

Issue #4, opened 2020-06-24
Ability to change numbering system on connectors


cocide:
I'm adding some 3 pin phones connectors to my layout and tracking the pins numerically (as 1, 2, 3) is not very useful since the connector is a tip, ring, sleeve type. I know I could add T, R, S into the pinout section, but I also need to track the names of what they are used for (in this case it is Dot, Dash, and Ground).

End results being: [KX3 Harness.pdf](https://github.com/formatc1702/WireViz/files/4826341/KX3.Harness.pdf)

```yml
connectors:
  Key:
    type: Phone Connector
    subtype: male 3.5
    pins: [T, R, S]
    pinout: [Dot, Dash, Ground]

cables:
  W1:
    gauge: 24 AWG
    length: 0.2
    color_code: DIN
    wirecount: 3
    shield: true

connections:
  -
    - Key: [S,R,T]
    - W1: [1,2,3]
  -
    - Key: S
    - W1: s
```

formatc1702:
Implemented in #10  :) Thanks @kimmoli

Issue #3, opened 2020-06-24
Ability to specify twisted wires


s7726:
Add ability to specify, and visualize, twisted wires with appropriate call outs for twist specs.

SnowMB:
I made a quick mockup how this could be visually achieved with hacking a little bit on the intermediate files graphviz can produce. This is similar to the solution I proposed in #120 and in my branch https://github.com/SnowMB/WireViz/tree/banded_wires_pydot.

I switched the order of wire row and text row for the 2nd wire in the twisted pairs so I can draw them overlapping without touching the text. I added a few points to the splines that make up the wire inside of the table to produce the twisting effect. This could be tweaked a little bit more to make a better looking.

Actually twisting them (switching which wire is drawn in front for each crossing) is way harder because they have to be split in peaces and each section has to be drawn in a different order.

Graph:
![test](https://user-images.githubusercontent.com/7050959/96110525-451e2a00-0ee0-11eb-816e-9a5741e26d99.png)


Hacked file `tutorial02.gv`: 
```
graph G {
outputorder=nodesfirst;
rankdir=LR;
graph [bb="0,0,839,180", bgcolor=white, fontname=arial, nodesep="0.33", rankdir=LR, ranksep=2];
node [fillcolor=white, fontname=arial, label="\N", shape=record, style=filled];
edge [fontname=arial, style=bold];
X1 [height="1.9167", label=<<table border="0" cellspacing="0" cellpadding="0"><tr><td><table border="0" cellspacing="0" cellpadding="3" cellborder="1"><tr><td balign="left">X1</td></tr></table></td></tr><tr><td><table border="0" cellspacing="0" cellpadding="3" cellborder="1"><tr><td balign="left">Molex KK 254</td><td balign="left">female</td><td balign="left">4-pin</td></tr></table></td></tr><tr><td><table border="0" cellspacing="0" cellpadding="3" cellborder="1"><tr><td port="p1r">1</td></tr><tr><td port="p2r">2</td></tr><tr><td port="p3r">3</td></tr><tr><td port="p4r">4</td></tr></table></td></tr></table>>, margin=0, pos="91,83", shape=none, width="2.5278"];
[...]

PR #10, opened 2020-06-24
Add optional custom pinnumbers to connector

kimmoli:
One possible solution for custom pin numbering, e.g. fits needs in #4 

- Pincount is detected automatically if not given.
- If both, pinout and pinnumbers, are given, their count must match.
- Freedom to give pin numbers in any order, in alphabetical and mixed notation.

Sample

```
connectors:
  X1:
    type: D-Sub
    subtype: male
    pincount: 25
    pinnumbers: [1,14,3,16,5,18,7,20,9,22,11,24,13]
    pinout: [ SENSE_P_1, SENSE_N_1, SENSE_P_2, SENSE_N_2, SENSE_P_3, SENSE_N_3, SENSE_P_4,SENSE_N_4, SENSE_P_5, SENSE_N_5, SENSE_P_6, SENSE_N_6, GND ]
  X2:
    type: F48
    subtype: female
    pincount: 48
    pinnumbers: [ z2,b2,d2,z4,b4,d4,z6,b6,d6,z8,b8,d8,z10,b10,d10,z12,b12,d12,z14,b14,d14,z16,b16,d16,z18,b18,d18,z20,b20,d20,z22,b22,d22,z24,b24,d24,z26,b26,d26,z28,b28,d28,z30,b30,d30,z32,b32,d32 ]

cables:
  W1:
    gauge: 0.25 mm2
    length: 0.2
    color_code: DIN
    wirecount: 16
    shield: true

connections:
  -
    - X1: [1,14,3,16,5,18,7,20,9,22,11,24]
    - W1: [2,1,4,3,6,5,8,7,10,9,12,11]
    - X2: [d4,z2,d10,z8,d16,z14,d20,z18,d26,z24,d32,z30]
  -
    - X1: 13
    - W1: s

```
demo01

### DIFF
--- a/src/wireviz.py
+++ b/src/wireviz.py
@@ -82,16 +82,23 @@ class Harness:
                 # a = attributes
                 a = [n.type,
                      n.subtype,
-                     '{}-pin'.format(len(n.pinout)) if n.show_pincount else '']
+                     '{}-pin'.format(n.pincount) if n.show_pincount else '']
                 # p = pinout
                 p = [[],[],[]]
                 p[1] = list(n.pinout)
-                for i, x in enumerate(n.pinout, 1):
-                    if n.ports_left:
-                        p[0].append('<p{portno}l>{portno}'.format(portno=i))
-                    if n.ports_right:
-                        p[2].append('<p{portno}r>{portno}'.format(portno=i))
-                # l = label
+                if (n.pinnumbers):
+                    for i in n.pinnumbers:
+                        if n.ports_left:
+                            p[0].append('<p{portno}l>{portno}'.format(portno=i))
+                        if n.ports_right:
+                            p[2].append('<p{portno}r>{portno}'.format(portno=i))
+                else:
+                    for i, x in enumerate(n.pinout, 1):
+                        if n.ports_left:
+                            p[0].append('<p{portno}l>{portno}'.format(portno=i))
+                        if n.ports_right:
+                            p[2].append('<p{portno}r>{portno}'.format(portno=i))
+                    # l = label
                 l = [n.name if n.show_name else '', a, p, n.notes]
                 dot.node(k, label=nested(l))
 
@@ -318,6 +325,7 @@ class Connector:
     pincount: int = None
     notes: str = None
     pinout: List[Any] = field(default_factory=list)
+    pinnumbers: List[Any] = field(default_factory=list)
     color: str = None
     show_name: bool = True
     show_pincount: bool = True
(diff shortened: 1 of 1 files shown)


src/wireviz.py
```diff
@@ -327,15 +335,17 @@ def __post_init__(self):
...
+            if self.pinout:
                 self.pincount = len(self.pinout)
-        else:
-            if not self.pincount:
-                self.pincount = 1
-            self.pinout = ['',] * self.pincount
+            elif self.pinnumbers:
+                self.pincount = len(self.pinnumbers)
+            else:
+                raise Exception('You need to specify at least one, pincount, pinout or pincount')
```
elliotmr, line 344: I guess one of these should be pinnumbers

kimmoli: sure. fixed.

formatc1702:
This is a fantastic addition, I'll take care of it in the next few days!

formatc1702:
Unfortunately, this breaks the minimalist way of specifying only a pincount (e.g. Tutorial 01)

formatc1702:
Idea: Perhaps, instead of having `pinout` and `pinnumbers`, using `pin_labels` and `pin_descriptions` or something similar might be less confusing/ambiguous? Not sure if `pinnumbers` works when it's not only numerical labels...

kimmoli:
Thanks for your efforts on this project.

> 
> 
> Additional idea: Perhaps, instead of having `pinout` vs `pinnumbers`, using `pin_labels` or `pin_IDs` vs `pin_descriptions`, `pin_functions` or similar might be less confusing/ambiguous? Not sure if `pinnumbers` works when it's not only numerical labels...

IMHO `pinnumber` is pin number, regardles is it number, letter or mixed.

pinout could be `pinname`

i would avoid underscores ...

formatc1702:
`pinnumber` and `pinname` could work. I totally agree on the underscore thing. Thanks for helping out!

kvid:
IMHO `pinlabel` is more intuitive when the values might be non-numeric, and it works very well wirhout underscore.

kvid:
IMHO `pinnumbers` might be confusing when the values might be non-numeric. Why not use `pins` as suggested in issue#4 if these new pin identifiers will be the only way to reference pins? I would  also consider the possability to reference pins in alternative ways, like `#3` for the third pin and `$GND` for the pin named GND in the pinout, etc. The `pinlabels` and `pinnames` alternatives can also be considered without the underscore.

kvid:
I have not tested the proposed code to see if it is possible to specify `s` as pin identifier and how then a conflict with the shield identifier is handeled. Is this an issue?

PR #11, opened 2020-06-24
Add ability to specify part number

Tyler-Ward:
This looks like a useful tool already found some good uses for it, I needed to be able to add connector part numbers to one I used it for so added a field for them. Feel free to merge if you think this will be useful to others. example 02 has been updated to show an example of this.


formatc1702:
This looks great! I'll have to think about how to best handle part numbers... In my experience, it's helpful to have both a manufacturer part number, as well as an internal part number (for inventory, etc.), or at least the option to add both.

The idea has been on my mind for a while, as well, especially since reading the [Part Number Anthology eBook](https://blog.grabcad.com/wp-content/uploads/2015/01/A-Part-Number-Anthology-eBook.pdf) (see page 12 onwards).

kvid:
I suggest using `partnumber` to avoid the underscore, and I also suggest to include it in the BOM for wires in a bundle as well for completeness.

formatc1702:
ouch, you are probably correct. I guesss you would need to specify a list in that case... the question is how to handle repeated colors (and thus part numbers) in a bundle? would you need to specify the part number for every instance of that color in the bundle?

Tyler-Ward:
The part number anthology is a useful link. In which case its probably worth having fields for manufacturer, manufacturer part number and internal part number. with those it is probably worth formatting part number information into a separate line in the top info section.

Good spot on the bundle wires one way to avoid needing to specify it multiple times would be to use something like a wire info override field which allows for info on individual wires to be replaced. but adds some complexity to the format as a trade-off e.g.

```
templates:
  - &blueWire
    part_number: 12345

  W1:
    category: bundle
    length: 0.3
    colors: [BK, BU, BU, WH]
    wireinfo:
      1:
        part_number: 12344
      2: <<: &blueWire
      3: <<: &blueWire
      4:
        part_number: 12344
```

This could also be used to add support for other differing features of a wire in a bundle or cable such as gauge (but that probably wants to end up in its own issue).

kvid:
A straight foward implementation would need to repeat the numbers for repeated colors. To avoid that without moving wire properties into a new Wire class (a major refactoring) might end up as a bit ugly or user unfriendly solution, unless someone has a good idea. A refactoring like this will also open up for other differences between wires in a cable/bundle, like different gauge values (e.g. 2x1.5mm2+PE1mm2 cable), but it should then be raised as a new issue.

kvid:
@Tyler-Ward I did not see your last comment while I wrote mine.

Tyler-Ward:
What do people think of the following for displaying part numbers in the diagram.
ex02

currently I have it set up so the part numbers for bundles specified as lists until a better solution appears.

formatc1702:
That doesn't look too bad.

I'm assuming `X1`, `X2` and `X4` would also show `Company` and `IPN` information if it was specified?

Please see the [discussion](https://github.com/formatc1702/WireViz/pull/28#issuecomment-650716115) in #28 as well. I like the idea of having `Company` + `MPN` in one table cell, and `IPN` in another cell next to it.

Please also take BOM output into account. IMHO, empty columns should be deleted, e.g. if a hobbyist does not use internal part numbers, there's no use for an empty BOM column.

formatc1702:
Another thing to consider is how to render part numbers in ferrules, like in [example 6 from the tutorial](https://github.com/formatc1702/WireViz/blob/master/tutorial/readme.md#6---custom-ferrules).
I must admit that I have neglected ferrules a bit myself.

Tyler-Ward:
> I'm assuming `X1`, `X2` and `X4` would also show `Company` and `IPN` information if it was specified?

That is correct 

> Please also take BOM output into account. IMHO, empty columns should be deleted, e.g. if a hobbyist does not use internal part numbers, there's no use for an empty BOM column.

Each of the part number fields will only appear in the BOM if they have been used.
 
> Nit-pick: a space after the colon would make it look nicer: `MPN: 436400200`

Agreed. have added that.

I have now merged in the upstream refactoring changes, got the bom populating and added information for the ferrules.
image

```
connectors:
  X1: &boo
    type: Molex Micro-Fit
    subtype: male
    pinout: [GND, VCC]
    manufacturer_part_number: 436400200
  X2: &con_power_f # define template
    type: Molex Micro-Fit
    subtype: female
    pinout: [GND, VCC]
    manufacturer_part_number: 436450200
  X3:
    <<: *con_power_f # create from template
    manufacturer: Molex
    internal_part_number: ABC123

  X4:
    <<: *con_power_f # create from template

ferrules: # ferrules
  F1:
    type: Ferrule, crimp
    subtype: 0.5 mm²
    color: OG # optional color
    manufacturer: company
    manufacturer_part_number: FERR01
    internal_part_number: ABC123

cables:
    show_equiv: true
    length: 0.2
    manufacturer_part_number: 123456
  W2:
    <<: *wire_power # create from template
    category: bundle
    manufacturer_part_number: [123456,123457]
  W3:
    <<: *wire_power # create from template
    category: bundle
    manufacturer: [Company,Company]
    manufacturer_part_number: [123456,123457]
    internal_part_number: [CAB123,CAB456]

connections:
  -
    - X1: [1-2]
    - W1: [1-2]
    - X2: [1-2]
  -
    - X1: [1-2]
    - W2: [1-2]
    - X3: [1-2]
  -
    - X1: [1-2]
    - W3: [1-2]
    - X4: [1-2]
  -
    - F1
    - W1: [1,2]
```

formatc1702:
Looks great! Trying to coordinate merging this with #28 (see my comment there), looking forward to it.

formatc1702:
The only issue I have is, in the example, it doesn't make sense for `W1`'s cables to split both into`X1` and into the ferrules... I get that you did it to show it works, but it can probably be left out of the example.

formatc1702:
I realized your branch was based on the (relatively) old code before the refactoring (82b173f2cebe32b23c8df14bec9e0c2c299ee48c) and would not have played well with it.

For this reason, I have decided to merge #54 (updated version of #28) first... let's see if we can update this PR to work with the post-refactoring code. Sorry for the trouble.

formatc1702:
I'm working to merge this, including #54's nice use of lambdas (`connector_group`, `cable_group`) and integrating your new atttributes into them...

formatc1702:
There are some issues after the merge, will have a look tonight... Overall it looks promising.

formatc1702:
The merge seems to have been fixed. Both #11 and #54 are implemented.

* I fine-tuned the HTML output for the connector node to make the row with the part numbers visually consistent with the one for connectors (see [here](https://github.com/formatc1702/WireViz/blob/dev/tutorial/readme.md#8---part-numbers))
* FYI, you had directly edited `/tutorial/readme.md` and added your example. Instead, I created `tutorial08.md` and rebuild the examples so it wouldn't get lost.

Please let me know if any problems persist and submit them as new issues if necessary. Thanks a lot!

Tyler-Ward:
All looks good here, Thanks for fixing the html output for cables that one slipped through. Will open an issue with the discussion about extra wire info for later discussion.

PR #9, opened 2020-06-24
Refactor depency handling, installation, and add a few features

n42:
Hey,

Great idea and immediately useful, for me. I've done some quick and dirty - but fairly major - refactoring here as follows for my own sake - use it if you wish.

- Got rid of the graphviz and pyyaml dependencies that were included in the repo
- Add a setup.py script with some preliminary metadata and the required dependencies, in order to 'pip install' instead. It should be fairly easy to get it onto PyPi. 
- Updated documentation to reflect the above, rebuilt the demos
- Broke out the argument parsing to a separate function and removed the defunct 'test' default input (the file you pointed to wasn't included in the repo, so running without an argument crashed rather than printed the help)
- Feature: add a field for specifying exact cable SKUs and the like (_specified_cable_ key)
- Feature: add automatic AWG->mm2 conversion as well, when using _show_equiv_. Some minor refactoring there as well, modified ex06 to show it.

PR #17, opened 2020-06-25
Add support for Multicolored Wires

### LINKED ISSUE #12: [feature] Multiple colors per wire
Is it possible to have multiple colors per wire? If not possible yet, it would be a huge improvement. Thanks for all the hard work!

aakatz3:
Closed  #12, adding support for multicolored wires, along with example. Also extends the DIN color code, and adds the following color codes:

- T568A
- T568B
- even-count color code/25 pair color code/Telco color code


aakatz3:
GraphViz can make the wires banded in the stripes, however it causes half of the wire to be one color, the other half to be the other, and would require removing the black borders. I ended up adopting the twisted stripe commonly seen in CAT5/CAT6A cables.

For the gradient, It may be possible to have it add multiple "cells" that have the additional wire colors (this may be the best option so the thickness matches), but I thought it was likely better to have the gradient than to just have the primary color, so I left that default behavior in.


CAT6A

The gradient behavior happens due to line 152 of wireviz.py:
`bgcolor = ":".join(wv_colors.translate_color(x, 'hex').split(':')[:2])`

This line suppresses the warning about additional colors being ignored. To just have the major/first color, you can either change the 2 to a 1, or replace the line with the following:

`bgcolor = wv_colors.translate_color(x[:2], 'hex')`

src/wireviz.py
```diff
@@ -141,7 +142,7 @@ def create_graph(self):
                 for bla in p:
                     html = html + '<td>{}</td>'.format(bla)
                 html = html + '</tr>'
-                bgcolor = wv_colors.translate_color(x, 'hex')
+                bgcolor = ":".join(wv_colors.translate_color(x, 'hex').split(':')[:2])
                 html = html + '<tr><td colspan="{colspan}" cellpadding="0" height="6" bgcolor="{bgcolor}" border="2" sides="tb" port="{port}"></td></tr>'.format(colspan=len(p), bgcolor=bgcolor if bgcolor != '' else '#ffffff', port='w{}'.format(i))
```
kvid, line 146: I suggest changing height="6" to height="10" for two-colored wires to match the extra thickness of the connecting curved wires. The thickness seems to be 2x the number of color stripes (black edge + base color + stripe color + base color + black edge).

kvid: An even better improvement (to avoid the color gradient) is to build the striped wire using multiple HTML rows with the port attribute only present in the middle row for each wire. See this excerpt from the upper wire in ex07.gv where the black edges are separated rows:
<tr><td colspan="3" cellpadding="0" height="2" bgcolor="#000000" border="0" sides="tb"></td></tr>
<tr><td colspan="3" cellpadding="0" height="6" color="#ffffff" bgcolor="#00ff00" border="2" sides="tb" port="w1">
<tr><td colspan="3" cellpadding="0" height="2" bgcolor="#000000" border="0" sides="tb"></td></tr>
It might be easier to to use one row (with no border) for each color in the generic case with multiple stripes.

njhurst: SVG provides dashes ( https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray ) which provide a simple and clean way to add this.  Draw a solid wire then a dashed wire.

aakatz3: I just saw your changes, if you are able to, please check out my solution and let me knnow if you have any comments. I will try to review your pull request shortly.

kvid: I just created this PR: https://github.com/aakatz3/WireViz/pull/4 with my suggested implementation.

aakatz3: @njhurst While it would be nice to do this, there is a problem: Graphviz does not support this properly. When drawing dashed lines, you cannot specify an alternate color, and you can't seem to overlay them properly. One COULD modify the SVG after the fact, but it breaks compatibility with GraphViz

aakatz3:
@kvid I merged your pull request in, please let me know if it looks good

kvid:
Thank you. It looks good, except I found a minor issue in the odd case of a wire with more than two colors. See https://github.com/aakatz3/WireViz/pull/5

aakatz3:
Fixed and merged to my branch. Thank you!

X-Ryl669:
I think once there is a multicolored wire, all wires should be drawn with extended thickness, as it does not look very good with mixed thickness wires.

kvid:
@X-Ryl669 Do you prefer equal wire thickness within the same cable or globally in all cables? If we go in this direction, then maybe a difference in cable gauge also optionally could be visualized as a difference in wire thickness between different cables.

X-Ryl669:
Yes, it makes sense to have wire with same gauge using the same thickness (whether they are multi-colored or not).
Having wire with higher thickness for higher gauge is even better (but as I understand it currently, the gauge value is not interpreted, only used for mapping to AWG).

src/wireviz.py
formatc1702: This looks very promising!

> @njhurst While it would be nice to do this, there is a problem: Graphviz does not support this properly. When drawing dashed lines, you cannot specify an alternate color, and you can't seem to overlay them properly. One COULD modify the SVG after the fact, but it breaks compatibility with GraphViz

It is indeed possible to define [multi-colored dashed lines in GraphViz](https://graphviz.org/doc/info/attrs.html#d:color) (see example [here](https://edotor.net/?engine=dot#graph%20%7B%0A%0A%09edge%20%5Bcolor%3D%22yellow%3B0.2%3Agreen%3B0.1%3Ayellow%3B0.4%3Agreen%3Ayellow%3B0.2%22%5D%0A%0A%09A%20--%20B%0A%7D%0A)); however, it might be hard to make it look good, especially making it look even between the edges and the HTML table cell (which I guess could be subdivided to simulate stripes?).

formatc1702:
This looks very promising!

> @njhurst While it would be nice to do this, there is a problem: Graphviz does not support this properly. When drawing dashed lines, you cannot specify an alternate color, and you can't seem to overlay them properly. One COULD modify the SVG after the fact, but it breaks compatibility with GraphViz

It is indeed possible to define [multi-colored dashed lines in GraphViz](https://graphviz.org/doc/info/attrs.html#d:color) (see example [here](https://edotor.net/?engine=dot#graph%20%7B%0A%0A%09edge%20%5Bcolor%3D%22yellow%3B0.2%3Agreen%3B0.1%3Ayellow%3B0.4%3Agreen%3Ayellow%3B0.2%22%5D%0A%0A%09A%20--%20B%0A%7D%0A)); however, it might be hard to make it look good, especially making it look even between the edges and the HTML table cell (which I guess could be subdivided to simulate stripes?).

kvid:
The wire gauge value is not currently inerpreted, but one approach could be to sort all unique gauge values, use the minimum thickness for the smallest gauge value, add a constant to the thickness for the next larger gauge value, etc. If we need a maximum thicness limit, then the close gauge values can be grouped and use equal thickness. Alternatively, a less functional, but simpler implementation could be to add an optional wire thickness attribute in the YAML file.

formatc1702:
Perhaps it makes sense to split the wire gauge visualization into a separate issue, so as not to further complicate things. It would be a nice addition but is not necessarily related to banded/striped wires.

It will probably be necessary to allow for both lengthwise stripes (like Ethernet or PE) and color rings (like DIN 47100)... although lengthwise stripes seem to be easier to implement.

X-Ryl669:
Agreed. It's just that this current implementation enlarge the wire if it's multicolor but doesn't do that if it's monochrome.

aakatz3:
In order to make the dashed lines work properly in graphviz, they would need to lose the black outlines, and we would need a way to count and calculate the number of stripes that should be placed. While you can put two or more colors into multicolored lines, I couldn't find a way to pattern them without just repeating, over and over, the color code.

A possible solution to enlarging single color wires instead of just tripling up the color may be to change the stroke width, but I do not know if this can be done for only the center spline/stroke.

formatc1702:
I've recently made some rather big refactoring changes (#9), is it possible to rebase these changes onto the latest version to make merging easier?

aakatz3:
@formatc1702 I will try my best to do this ASAP. I will have to check out the new code, and will provide an update when I can.

aakatz3:
@formatc1702 The "bug" with varying wire thicknesses still persists, however I have merged my code into the refactored codebase. Please let me know if my changes have messed up any other code.

src/wireviz/Harness.py
```diff
@@ -39,7 +39,7 @@ def create_graph(self):
         font = 'arial'
         dot.attr('graph', rankdir='LR',
                  ranksep='2',
-                 bgcolor='white',
+                 bgcolor=wv_colors.default_bknd_color,
```
formatc1702, line 42: default_bg_color is more intuitive IMHO

src/wireviz/wv_colors.py
```diff
@@ -1,13 +1,37 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
+import sys
 
 COLOR_CODES = {
-    'DIN': ['WH', 'BN', 'GN', 'YE', 'GY', 'PK', 'BU', 'RD', 'BK', 'VT'], # ,'GYPK','RDBU','WHGN','BNGN','WHYE','YEBN','WHGY','GYBN','WHPK','PKBN'],
+    'DIN': ['WH', 'BN', 'GN', 'YE', 'GY', 'PK', 'BU', 'RD', 'BK', 'VT', 'GYPK', 'RDBU', 'WHGN', 'BNGN', 'WHYE', 'YEBN',
```
formatc1702, line 6: Thanks for your effort 🥇

src/wireviz/wv_colors.py
```diff
@@ -19,10 +43,17 @@
     'TQ': '#00ffff',
     'BU': '#0066ff',
     'VT': '#8000ff',
-    'BN': '#666600',
+    'BN': '#a52a2a',
```
formatc1702, line 46: Not a huge fan of the new brown color, it looks too similar to orange, see example 01

Screen Shot 2020-07-01 at 21 03 17

src/wireviz/wv_colors.py
```diff
@@ -50,29 +85,52 @@
...
     'VT': 'vi',
     'BN': 'br',
+    # To the best of my ability, likely incorrect:
+
+    # Slate --> Schieferfarbe --> SI ??
+    'SL': 'si',
+    # Copper
+    'CU': 'ku',
+    # Tinned
+    'SN': 'si'
```
formatc1702, line 95: should be 'zn' (Zinn) probably

formatc1702:
Also, maybe exclude the built examples from the merge to reduce clutter... the input YAML files should be sufficient. I will simply rebuild them before I merge into master. Thanks a lot for your work (everyone involved)!

.gitignore
```diff
@@ -8,3 +8,21 @@ build
 data
 dist
 venv/
+
+.vscode/
+
+
+# Jetbrains Stuff
```
formatc1702, line 15: Not sure if all these IDE and Windows lines should be included in the repo .gitignore... they should be ignore locally IMHO

formatc1702:
@aakatz3 please let me know if you are able to include the suggested changes. If not, no worries, I can take care of it as well, I'd just like to get this merged soon since it's an important, but maybe not so trivial change :)

aakatz3:
Do you have a suggestion for a brown color? I made a mistake, but the old brown looked close to one of the other colors. I made a pallate of all the colors, which can be seen here:
[colorscheme(1)](https://colourco.de/freebuild/aaaaaa-84878c-fffbf8-000000-ffffff-999999-ff66cc-ff0000-ff8000-ffff00-00ff00-00ffff-0066ff-8000ff-a52a2a-708090-d6775e-895956-54aa85-f7f854-666600)

The old brown is the last color in the palate

Maybe [this](https://www.pantone.com/color-finder/724-XGC), or the second copper color would be a good alternative?

Since the wires don't do gradients anymore, I may as well consider scrapping the copper and tinned copper alternate colors

I re-calibrated my display and see what you mean.


On a separate note, what dev enviornment/IDE are you using?

.gitignore
aakatz3: MacOS ones are ATM, so I was putting the windows set (from Github's recommendations) in just in case. I don't necessarily see any harm. I can try to think if there is a better way, probably within subfolders. I will also clean up the Jetbrains stuff.

formatc1702:
> Do you have a suggestion for a brown color?

LOL, I had literally already pasted the link to the Technology Connections video in my response before re-reading your comment and finding it there. Maybe [saddle brown (#8B4513)](https://www.w3schools.com/colors/color_tryit.asp?color=SaddleBrown)? Quite similar to your suggestion. Or the one from the wiki link below.

> Since the wires don't do gradients anymore, I may as well consider scrapping the copper and tinned copper alternate colors

It was a fancy idea, but yes, one color for each material should be enough. Maybe include an option for gold (GD) and silver (SR) as well while we are at it? [Wikipedia](https://en.wikipedia.org/wiki/Electronic_color_code#Color_band_system) might serve as an inspiration.

> On a separate note, what dev enviornment/IDE are you using?

I'm using Atom as my text editor, and [Fork](https://git-fork.com/) as my Git GUI (my visually oriented brain refuses to do Git via CLI).

.gitignore
formatc1702: Point taken for the windows stuff! :+1:

src/wireviz/wv_colors.py
```diff
@@ -19,10 +43,17 @@
...
     'VT': '#8000ff',
-    'BN': '#666600',
+    'BN': '#a52a2a',
+    'SL': '#708090',
+    # Faux-copper look, for bare CU wire
+    'CU': '#d6775e:#895956',
+    # Silvery look for tinned bare wire
+    'SN': '#aaaaaa:#84878c',
+    # Yellow-green PE wire
+    'PE': '#54aa85:#f7f854:#54aa85',
```
formatc1702, line 53: Do we need separate yellow and green tones for PE? I'd rather use the existing YE and GN

aakatz3: I can change those to the same. To be honest I kinda just grabbed the yellow and green for it from a random standards document. I am going to make a separate issue to standardize the wire colors, so that others can comment if necessary.  I'm going to make the file-level changes to the repo, then start fixing colors.

formatc1702:
@aakatz3 could you give an estimate of when this will be ready to merge?
I have a lot of features and bugs I want to include, but I'd like for them to build on top of this PR.
If you are unable to address it soon, no worries, me or someone else ca take over.
Thanks!

kvid:
@aakatz3 do you need more information from me about my last comment about UTF-8 encoding differences? I hope this minor issue is not the thing that stop this important PR from being merged.

I can also contribute about equalizing the wire thickness if you want me to, but I suggest merging the main part first, and then we can create a new PR for the thickness issue.

formatc1702:
> I can also contribute about equalizing the wire thickness if you want me to, but I suggest merging the main part first, and then we can create a new PR for the thickness issue.

Yes please, submit any new features as new PRs afterwards.

formatc1702:
There hasn't been any activity in the last 10 days, but I really would like to merge this ASAP.

@aakatz3 any news?
@kvid do you have the time to take over this issue, look over the requested changes and get this ready to merge into the current `dev` branch with no merge conflicts?

Only addendum to the code comments: Ideally please revert any changes to the example output files, keeping only the input .yml files, so the diff of the commit is not so big. I will rebuild and publish all examples when we merge into master.

src/wireviz/wv_colors.py
formatc1702: please remove PE since it's not a color. it should be implemented as GNYE / YEGN like any other banded/striped wire.

.gitignore
formatc1702: Please keep the Windows stuff (`thumbs.db`, `desktop.ini`) but clear the VSCode/JetBrains stuff since it's too specific.

kvid:
I'm a bit busy at day-time these days, but I can put in some effort at evenings to get this PR ready. Unless @aakatz3 suddenly gets back, I suggest I clone his branch [aakatz3:feature-multicolor-wires](https://github.com/aakatz3/WireViz/tree/feature-multicolor-wires) and address the change requests and merge conflicts. Then I can create a new PR with my version of the branch referring to this one. This way @aakatz3 or someone else can continue this PR in case I fail or use too much time trying.

formatc1702:
Thanks, appreciate it. See if you can rebase your `feature` branch onto your `dev`, I want to avoid having a tangle in the commit tree in the end :)

aakatz3:
I've been working on trying to rebase my commit tree, but I keep running into issues within the HTML, so I was trying to look into refactoring the HTML. I don't quite know the best way forward. I may do a squash rebase and submit a new pull request. Do you have any thoughts on the best way to tackle this?


On a separate note, I found the problem with UTF-8, encodings weren't hardcoded or detected, so on windows, it would write utf-8 but read as codepage2512, which caused the issue (or vice versa). My personal git tree is a major mess at this point trying to figure this out.

formatc1702:
@aakatz3 glad you're back!

Honestly, so much has changed since work on this issue started, my naïve approach would be to stat from the current state of the `dev` branch, and manually build in the changes there.
No sense in desperately trying to untangle the commit tree! Not sure if the other co-authors take issue with removing the development history of this feature, but pragmatically speaking, it would be the easiest.

formatc1702:
This might mean starting a clean branch from `dev`, porting your changes to there, and submitting a new PR. Let me know if you want to move forward with this.

kvid:
What are the major differences between your last pushed commit and your personal commits? If you push your last commits in a separate temporary branch, then I can have a look and perhaps suggest a way forward.

kvid:
I am thinking it's better to merge than to rebase in this case to avoid a lot of non-trivial conflicts because files have been moved in both branches. I sugggest creating a branch before `Merge branch 'dev' into feature-multicolor-wires` to reduce the tangles.
I tried this in my personal repo:
```
git branch aakatz3-multicolor e6317536f75ef5ed45dde0133c973a13145d704f
git checkout dev
git merge aakatz3-multicolor
```
I had to resolve these conflicts:
- .gitignore   (keeping changes from both branches)
- src/wireviz.py   (deleting file)
- src/wireviz/build_examples.py   (keeping the highest number of examples and tutorials)
- src/wv_colors.py   (deleting file)

After this merger, I could build all examples without errors, but I have not yet checked if all changes was merged in correctly.

formatc1702:
> I am thinking it's better to merge than to rebase in this case to avoid a lot of non-trivial conflicts because files have been moved in both branches.

That's precisely why I am advocating for a fresh start from `dev`, cleanly re-implementing the desired changes and getting rid of the messy history.

aakatz3:
Dev-reset complete, please see #96

kvid:
I did rebase `aakatz3-multicolor` described above onto `dev` and adapted the commits to use f-strings:
https://github.com/kvid/WireViz/tree/aakatz3-multicolor-rebased
Then I discovered the new PR #96 from @aakatz3 and stopped. It was a nice excercise though. Each commit is verified by running the build script and inspecting some critical diagrams. It would be fun to comoare this with the commits in PR #96 if I have the time. :-)

formatc1702:
I've merged #96 into my own feature branch, will tweak and merge into `dev` soon. Closing this one. Thanks everyone for the hard work.

PR #28, opened 2020-06-28
Support named type of cables (or wires in a bundle)

kvid:
The `type` attribute is already accepted without errors if used for cables, but as far as I can see, its value is currently not used in any output. This suggested code change show any cable type value (e.g. product number and/or name) below the cable name in the diagram and in the cable or wire entries of the BOM. The value is also included in the internal set of values that must be equal to share the BOM entry.

Please let me know if this is a wanted feature, and if so, suggest what should be added or changed before it can be merged, e.g. how many examples I should modify or add to demonstrate the new feature.

### DIFF
--- a/src/wireviz.py
+++ b/src/wireviz.py
@@ -128,6 +128,9 @@ class Harness:
             html = html + '<table border="0" cellspacing="0" cellpadding="3" cellborder="1">' # name+attributes table
             if c.show_name:
                 html = html + '<tr><td colspan="{colspan}">{name}</td></tr>'.format(colspan=len(a), name=c.name)
+            if c.type: # If the cable (or wires in a bundle) has a named type (e.g. product number+name)?
+                # TODO: If product number becomes a separate attribute, it could fit into the same table row.
+                html = html + '<tr><td colspan="{colspan}">{type}</td></tr>'.format(colspan=len(a), type=c.type)
             html = html + '<tr>' # attribute row
             for attrib in a:
                 html = html + '<td>{attrib}</td>'.format(attrib=attrib)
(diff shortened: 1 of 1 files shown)


formatc1702:
This is a nice idea, and I like the implementation.
I think this could complement #11 nicely; perhaps `type` could share the existing row with `wirecount`, `gauge`, etc.
In addition, both cables and connectors could optionally have a new row showing `partno_int` and `partno_ext` (see [my comment](https://github.com/formatc1702/WireViz/pull/11#issuecomment-650520043) in #11). I'm open for other naming conventions as well.

kvid:
I was afraid a type value in addition to the existing columns would make the frame too wide in many cases, but I see your point of beeing consistent with connectors.

formatc1702:
I've recently made some rather big refactoring changes (#9), please use the updated version to implement this to avoid any ugly merge conflicts... thanks!

kvid:
It was hard to merge the dev branch int my branch after the major refactoring. My solution was to reimplement the same functionality based on the last release. To make my job a bit easier, I first made some generic changes of the BOM generation code. These changes will probably also be useful for https://github.com/formatc1702/WireViz/pull/11 that have some common issues. I have pushed my changes to https://github.com/kvid/WireViz/tree/feature/cable_type_rev2 and included a couple of changes to the examples to demonstrate the new functionality, but please suggest more realistic cable type values. Is it possible to update this PR, or is it easier to create a new one?

formatc1702:
It's good that you have an eye on #11, since they are quite tightly linked.

Since it seems that #11 is ready to be merged without conflicts, what do you think about the following?

I will merge #11 into `dev` first, and you can merge it into your branch as well.
Then #28 (this) can be merged into `dev` afterwards, having been tested by you to work together with #11.

> Is it possible to update this PR, or is it easier to create a new one?
It's probably possible to update this PR; unfortunately, I lack experience to give exact instructions... 😐

kvid:
I believe https://github.com/formatc1702/WireViz/pull/54 is already based om dev. Did you have a look. I think it is ready,  except maybe for a better cable type value in the examples.

formatc1702:
Merged in #54

PR #30, opened 2020-06-28
Remove 3rd party libraries, add requirements.txt, and update readme

Cabalist:
Very neat project!

I removed the 3rd party libraries that are in this repo.  

NOTE: I didn't see the #9 PR until after I'd already done this but this fix is a self contained version of just one part of that one.   :)


formatc1702:
Thanks! I've integrated #9 's refactoring which also removes 3rd party libraries, so I'll mark this one as closed.

I'm not sure what the practical difference is between having the dependencies listed explicitly in requirements.txt vs having them auto-install in setup.py (as in #9); feel free to enlighten me :)

PR #34, opened 2020-06-28
[ci] Add Github Actions

flopp:
This add a Github Actions workflow which 
- sets up an Ubuntu image with graphviz & Python 3.7 + 3.8
- runs `build_examples.py` as an integration test
- is executed on each pull/pull request

The workflow can be extended to run some helpful analysis tools, e.g. `pylint` (Python linter), `mypy` (Pyhton type checking), `black` (format checker).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,25 @@
+name: Create Examples
+
+on: [push, pull_request]
+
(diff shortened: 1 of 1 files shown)


formatc1702:
Thanks! GitHub Actions is new to me, will give it a try.

PR #37, opened 2020-06-28
Fix for Designators apearing across all colours of wires of a guage.

### LINKED ISSUE #33: [bug] incorrect designators for different coloured wires in bundles
When producing a bom containing several bundles of wires with different colours the designators are added to all wires for that gauge rather than just the ones that colour appears in. Below is a simple example based on ex02.yml. Will likely look at fixing this later today but recording it here in the meantime.

```
connectors:
  X1: &boo
    type: Molex Micro-Fit
    subtype: male
    pinout: [GND, VCC]
  X2: &con_power_f # define template
    type: Molex Micro-Fit
    subtype: female
    pinout: [GND, VCC]
  X3:
    <<: *con_power_f # create from template
  X4:
    <<: *con_power_f # create from template

cables:
    category: bundle
    show_equiv: true
    length: 0.2
  W2:
    <<: *wire_power # create from template
    colors: [BU, WH] # number of wires implicit in color list
  W3:
    <<: *wire_power # create from template

connections:
  -
    - X1: [1-2]
    - W1: [1-2]
    - X2: [1-2]
  -
    - X1: [1-2]
    - W2: [1-2]
    - X3: [1-2]
  -
    - X1: [1-2]
    - W3: [1-2]
    - X4: [1-2]
```

image

Tyler-Ward:
This fixes #33 where designators appear across all wires of a guage rather than just the ones that use that colour.

The result of the testcase in that issue can be seen below.
image

### DIFF
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -287,15 +287,13 @@ class Harness:
                 for bundle in items.values():
                     # add each wire from each bundle to the wirelist
                     for color in bundle.colors:
-                        wirelist.append({'gauge': shared.gauge, 'gauge_unit': shared.gauge_unit, 'length': shared.length, 'color': color, 'designators': list(items.keys())})
+                        wirelist.append({'gauge': shared.gauge, 'gauge_unit': shared.gauge_unit, 'length': shared.length, 'color': color, 'designator': bundle.name})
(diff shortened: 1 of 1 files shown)


formatc1702:
Fantastic! Thanks.

PR #41, opened 2020-06-29
Feature: bidirectional AWG/mm2 unit conversion

### LINKED ISSUE #40: [Feature] Bidirectional awg<->mm2 unit conversion
Currently only mm2 -> awg conversion is handled, but the opposite is used/needed just as often, in my experience.

n42:
Broke out this feature from PR #9 and resubmitting separately with minor changes after rebase:

- Feature: Add automatic AWG->mm2 conversion as well, when using _show_equiv_. Some minor refactoring there as well, modified ex02 to show it the results. Note that the lookup table has been changed to strings on both sides, this is because e.g. '000', '2/0' and the like are perfectly valid AWG values.

- Fixes #40


formatc1702:
This looks great. Can we have AWG be output in upper case, and have it detect upper and lower case in the YAML input? It's the only nit-pick I can come up with ;)

n42:
Hmm the converted value is (hardcoded) upper case, the lowercase you see is from the original YAML; that's passed directly into the _gauge_unit_ variable and used as is. To clarify, do you mean that 1) all AWG strings should be converted to upper case for rendering, or that 2), all AWG strings should respect and use the casing from the YAML file? The first i think should be separately refactored, but it would yield more consistent output.

formatc1702:
Your edit beat me to the punch.

I think AWG should always be rendered as upper case, but we _do_ need to take into account where unknown units ('bananas', square inches, ...) are used, these should simply be passed through, with no equivalent conversion and no case conversion either.

n42:
Okay, so for the purposes of this PR:
- retain mm2 -> awg conversion, label as 'AWG'
- add check for gauge_type == awg before _othe_ -> mm2 conversion
- leave anything else as is even if show_equiv == true
- separate refactor PR for upper-i-fying all AWG string occurances

sound about right?

formatc1702:
Sounds good. If there's a separate PR coming for the last point, is this one ready for merging?

n42:
Actually that was a quick fix, pushing a commit with that as well in a sec seeing as it's related

src/wireviz/Harness.py
```diff
@@ -113,9 +113,17 @@ def create_graph(self):
                                  f'{connector.name}:p{loop[1]}{loop_side}:{loop_dir}')
 
         for _, cable in self.cables.items():
-            awg_fmt = f' ({awg_equiv(cable.gauge)} AWG)' if cable.gauge_unit == 'mm\u00B2' and cable.show_equiv else ''
+
+            if cable.show_equiv:
+                if cable.gauge_unit =='mm\u00B2':
+                    awg_fmt = f' ({awg_equiv(cable.gauge)} AWG)'
+                else:
```
kvid: Should this be `elif cable.gauge_unit.upper() == 'AWG' or cable.gauge_unit == ''` or something similar to avoid converting 'bananas' to AWG by assuming 'bananas' equals mm2?

n42:
Figured that might as well be done early-on, so put it alongside the mm2 -> mm^2 conversion in the dataclass

src/wireviz/Harness.py
kvid: The 'AWG' and 'mm' constant strings should also be defined only once in the helper file and then use identifiers in the rest of the code.

formatc1702: 👍  especially the Unicode escaping is a pain. Will you include the fix in this PR? should be easy enough

n42: yeah, that was fixed in 625b21e. And i agree about the multiple definitions, but that's a separate refactor job IMO

formatc1702: Up to you. Let me know when you're ready to merge this one.

n42: If you're fine with the string casing fix for now i'd say go for merge

kvid: @n42 I'm sorry for pointing out the issue you already had fixed, but I could not see your fix at that time. Github have been slow and somewhat unstable today, so I guess my view was not fully updated.

PR #44, opened 2020-06-29
Changed url in setup.py to point to original repo

Jannibal:


Jannibal:
I noticed that this had been fixed in a branch. Closing

PR #54, opened 2020-07-02
[feature] Support cable type rev2

kvid:


kvid:
See https://github.com/formatc1702/WireViz/pull/28

formatc1702:
Thanks so much

kvid:
Thank you for for accepting my contribution now,  and doing the merger with #11. I am on vacation now and cannot put in so much effort this week. I was thinking cable type values could be, e.g. Cat5e cable or PN wire, but did not find a matching example. However, that can be changed later.

PR #58, opened 2020-07-04
Update requirements.txt

gstein:
The "graphviz" Python module is required. Clarify that within the requirements.

(the graphviz/dot commands are needed at the OS level, but that can't be fixed here)


formatc1702:
What's the difference between declaring this in `requirements.txt` as opposed to [`setup.py`](https://github.com/formatc1702/WireViz/blob/master/setup.py#L23)?

Since I'm not familiar with creating Python packages myself, maybe @n42 can be of help as the author of the refactoring?

gstein:
That's quite a fair question, and I see that pyyaml is in `setup.py` (which could/should also be moved!) ... by moving it to `requirements.txt`, tools can consume the set of needed modules (whereas, they cannot discover the set from setup.py).

For example:
`$ pip install -r requirements.txt`

gstein:
(honestly, I didn't even stop to look into setup.py ... I always go to requirements.txt first :p )

n42:
A difference is iirc that requirements.txt is mainly useful in that it's easily human-readable and can be used for explicitly installing only the dependencies, which might be useful for using packages without installing them first. Pip/setuptools will use what's listed in setup.py as required dependencies and install them automatically if you use that to install the package.

n42:
It doesn't hurt anything to use both methods, but I'd suggest adding the pyyaml dep in the reqs file as well if you want to use that.

formatc1702:
So it's fine to have in both places? I'll just add two lines into `requirements.txt` file and leave `setup.py` untouched. Thanks.

n42:
Yeah, perfectly fine. There should be a way to have setup.py parse requirements.txt instead of having multiple places to define them as well, but I seem to recall something about pip version fuckery wrt that. I'll look into it when I'm at an actual computer.

formatc1702:
Cheers! I'll wait until then, I don't see this as a super urgent issue.

gstein:
Agree: not urgent. Per discussion, I've included `pyyaml` into the PR.

formatc1702:
Merged and closed for now. If we want to update `setup.py` to parse `requirements.txt` in the future, we can open a new PR. In the meantime, if more requirements appear, they'll be added to both files.

Issue #56, opened 2020-07-02
[feature] More control over wire parameters


Tyler-Ward:
Based on discussions in #11  it might be worth adding the ability alter more properties of individual wires in a cable or bundle e.g. guage (an example would be a usb cable with two thinner wires and two power wires). This will obviously add some additional complexity to both the definition format and to rendering cables/bundles with multiple parameters.

This feature could also allow for sub wires to be templated which will make reusing wires in multiple bundles or several times in the same bundle easier. A modified format concept from  #11 is shown below for reference.
``` yaml
templates:
  - &blueWire
    part_number: 12345
    color: BU

  W1:
    category: bundle
    length: 0.3
    wireinfo:
      1:
        part_number: 12344
        color: BK
      2: <<: &blueWire
      3: <<: &blueWire
      4:
        part_number: 12343
        color: WH
```

formatc1702:
A simple solution might be to allow every parameter of the cable/bundle to be either a string / int (applying it to all wires) or a list (length = wirecount),similar to how e.g. part numbers for bundles are handled now. It would be slightly more concise, and, when formatted cleverly, easier to read?

```yaml
W1:
  category: bundle
  length: 0.3 # maybe even this could become a list? does it make sense?
  colors:                   [BK,    BU,    BU,    RD   ]
  gauge:                    [0.25,  0.25,  0.5,   0.5  ]
  manufacturer_part_number: [12344, 12345, 12346, 12343]
```

Templating would not be as easy, though. Just a thought.

tim292stro:
Parameterizing is one of the major convenience factors of tools found in the commercial marketplace.  For example being able to say "This part" then have that part be defined by a separate drawing, which is then pulled in at compile time - allows for that to be a separate file even that can be managed separately.

Similarly, being able to do math in line on a parameter, allows for cases like "this branch needs to be 2/3 of the way along this trunk", where the trunk length is a variable and the Branch(Length) = Trunk(Length) * (2/3).  This way if the trunk length is changed, the branch length will be changed maintaining the ratio.

I think since this can become multi-dimensional, it's a good ideal to break this out earlier in a file as a general practice - much like doing "global defines".  When it gets down to a bundle, this would be an array or arrays, with each conductor definition being its own array of defined attributes inherited from the conductor definition.

tim292stro:
I think this is also a suitable feature request to piggyback with another suggestion - that wire parameters not be limited to simply DC electrical characteristics.  Consider this type of cable:
[...]

PR #59, opened 2020-07-04
Simplify and fix bug in BOM generation

formatc1702:
@Tyler-Ward please have a quick look if your comments are addressed and I will merge into `dev`.

I can confirm all BOM outputs from the examples and tutorial remain unchanged when building with the new code.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -311,23 +311,19 @@ class Harness:
             name = f'Connector{conn_type}{conn_subtype}{conn_pincount}{conn_color}'
             item = {'item': name, 'qty': len(designators), 'unit': '',
                     'designators': designators if shared.category != 'ferrule' else ''}
-            if shared.manufacturer is not None:  # set manufacturer only if it exists
-                item['manufacturer'] = shared.manufacturer
-            if shared.manufacturer_part_number is not None:  # set part number only if it exists
-                item['manufacturer part number'] = shared.manufacturer_part_number
-            if shared.internal_part_number is not None:  # set part number only if it exists
-                item['internal part number'] = shared.internal_part_number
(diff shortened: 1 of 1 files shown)


src/wireviz/Harness.py
```diff
@@ -336,66 +326,54 @@ def bom(self):
...
-                item['manufacturer part number'] = shared.manufacturer_part_number
-            if shared.internal_part_number is not None:  # set part number only if it exists
-                item['internal part number'] = shared.internal_part_number
+            item = {'item': name, 'qty': round(total_length, 3), 'unit': 'm', 'designators': designators,
+                    'manufacturer': shared.manufacturer, 'manufacturer part number': shared.manufacturer_part_number, 'internal part number': shared.internal_part_number}
             bom_cables.append(item)
         # bundles (ignores wirecount)
         wirelist = []
         # list all cables again, since bundles are represented as wires internally, with the category='bundle' set
-        bundle_group = lambda b: (b.category, b.type, b.gauge, b.gauge_unit, b.length) # TODO: Why is b.length included?
```
formatc1702, line 353: @kvid I think this was your question... right now I guess it's not important to have `length` included in the check, but see #56 , potentially this might become relevant in the future.

Tyler-Ward:
That looks good to me, much cleaner than the way i originally did it. I have tested it with the files I used for testing the part numbers and they all generated fine.

src/wireviz/Harness.py
kvid: As long as the wires can be grouped across bundles, I don't see the need to group bundles at all. If needing the option to have cut wires in the BOM instead of total length, then that will affect only the wire grouping, as far as I can see. I'm planning a PR with such a simplification.

PR #62, opened 2020-07-05
BOM simplifications

kvid:
Three suggested independent commits that affect generation of BOM:

1. Correct a wrong TODO statement.

2. Remove bundle grouping as it is not needed for the wire grouping. 
As long as the wires can be grouped across bundles, 
grouping the bundles has no purpose, as the same
wires can be collected by just looping all bundles. 

3. Allow one common value or list of values for each wire. 
Allow one common value when they are equal for all wires, 
e.g. the manufacturer is often equal for all wires, 
or a list of values to allow each wire a different value.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -307,8 +307,7 @@ class Harness:
         bom_cables = []
         # connectors
         connector_group = lambda c: (c.type, c.subtype, c.pincount, c.manufacturer, c.manufacturer_part_number, c.internal_part_number)
-        groups = Counter([connector_group(v) for v in self.connectors.values()])
-        for group in groups:
+        for group in Counter([connector_group(v) for v in self.connectors.values()]):
             items = {k: v for k, v in self.connectors.items() if connector_group(v) == group}
             shared = next(iter(items.values()))
             designators = list(items.keys())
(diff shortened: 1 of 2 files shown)


src/wireviz/Harness.py
```diff
@@ -9,6 +9,11 @@
 from typing import List
 
 
+# Return the value indexed if it is a list, or simply the value otherwise.
+# TODO: Maybe move this to wv_helper?
```
formatc1702, line 13: Agree. Move to `vw_helper`.

src/wireviz/Harness.py
```diff
@@ -344,18 +349,15 @@ def bom(self):
         # bundles (ignores wirecount)
         wirelist = []
         # list all cables again, since bundles are represented as wires internally, with the category='bundle' set
-        bundle_group = lambda b: (b.category, b.type, b.gauge, b.gauge_unit, b.length)
-        groups = Counter([bundle_group(v) for v in self.cables.values() if v.category == 'bundle'])
-        for group in groups:
-            items = {k: v for k, v in self.cables.items() if bundle_group(v) == group}
-            shared = next(iter(items.values()))
-            for bundle in items.values():
+        # TODO: Maybe rename bundle to cable or maybe loop only [c for c in self.cables.values() if c.category == 'bundle']
```
formatc1702, line 352: I think bundle is fine, since that is what we are looking for. If becomes clear after `if bundle.category == 'bundle':` at the latest.

formatc1702:
I like this change! Simplification is always good. See my comments above.. I'll propably squash the commits into one once you're done to keep it simple, hope that's OK.

PR #64, opened 2020-07-05
Add support for line breaks in certain fields

### LINKED ISSUE #49: [feature] Add support for line breaks
Would be great to be able to add line breaks into the fields as long descriptions can extend the output past normal screen width

formatc1702:
Closes #49. @Tyler-Ward perhaps you can can have a quick look and let me know if something's missing.

I am aware that ferrules are not rendered very nicely if their fields include line breaks (try my [test code](https://gist.github.com/formatc1702/531e68e3da2c180006808acdaf797cfe)) so I'll tackle that next. It's a cosmetic issue not related to the actual line break support.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -343,7 +343,7 @@ class Harness:
             if bundle.category == 'bundle':
                 # add each wire from each bundle to the wirelist
                 for index, color in enumerate(bundle.colors, 0):
-                    wirelist.append({'gauge': bundle.gauge, 'gauge_unit': bundle.gauge_unit, 'length': bundle.length, 'color': color, 'designator': bundle.name,
+                    wirelist.append({'type': bundle.type, 'gauge': bundle.gauge, 'gauge_unit': bundle.gauge_unit, 'length': bundle.length, 'color': color, 'designator': bundle.name,
                                      'manufacturer': index_if_list(bundle.manufacturer, index),
                                      'manufacturer part number': index_if_list(bundle.manufacturer_part_number, index),
                                      'internal part number': index_if_list(bundle.internal_part_number, index)})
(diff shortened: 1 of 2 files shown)


formatc1702:
The graphical output looks good IMO.

test

src/wireviz/Harness.py
```diff
@@ -356,8 +356,8 @@ def bom(self):
             designators = list(dict.fromkeys(designators))  # remove duplicates
             designators.sort()
             total_length = sum(i['length'] for i in items)
-            wire_type = f', {remove_line_breaks(shared["type"])}' if 'type' in shared else ''
-            gauge_name = f', {shared["gauge"]} {shared["gauge_unit"]}' if 'gauge' in shared else ''
+            wire_type = f', {remove_line_breaks(shared["type"])}' if shared.get('type', None) else ''
+            gauge_name = f', {shared["gauge"]} {shared["gauge_unit"]}' if shared.get('gauge', None) else ''
```
formatc1702, line 360: The new syntax catches both _missing_ keys and keys with `None` as value.

src/wireviz/Harness.py
```diff
@@ -343,7 +343,7 @@ def bom(self):
             if bundle.category == 'bundle':
                 # add each wire from each bundle to the wirelist
                 for index, color in enumerate(bundle.colors, 0):
-                    wirelist.append({'gauge': bundle.gauge, 'gauge_unit': bundle.gauge_unit, 'length': bundle.length, 'color': color, 'designator': bundle.name,
+                    wirelist.append({'type': bundle.type, 'gauge': bundle.gauge, 'gauge_unit': bundle.gauge_unit, 'length': bundle.length, 'color': color, 'designator': bundle.name,
```
formatc1702, line 346: We were ignoring the type for bundles. In general, I see no reason to exclude it here.

Tyler-Ward:
This looks good to me, I think you have got all the fields that make sense to add line-breaks to and i couldn't spot any places those fields are printed which doesn't go through one of the three line break functions.

Issue #48, opened 2020-06-30
[feature] How to short multiple pins on one connector?


azivkovic3l:

formatc1702:
There is primitive functionality to attach two pins of the same connector. Please see the end of this modified example:

```yaml
connectors:
  X1:
    type: D-Sub
    subtype: female
    pinout: [DCD, RX, TX, DTR, GND, DSR, RTS, CTS, RI]
  X2:
    type: Molex KK 254
    subtype: female
    pinout: [GND, RX, TX]

cables:
  W1:
    gauge: 0.25 mm2
    length: 0.2
    color_code: DIN
    wirecount: 3
    shield: true

connections:
  -
    - X1: [5,2,3]
    - W1: [1,2,3]
    - X2: [1,3,2]
  -
    - X1: 5
    - W1: s
  - # short pins 6,7,8 together
    - X1: [6,7]
    - X1: [7,8]
```

See the output below. It's not pretty, so I am open for suggestions on how to best visualize this. I guess there is also a difference between running a wire between two pins, and connectors that allow some other sort of shorting internally.

![test](https://user-images.githubusercontent.com/2447251/86153338-cef8e180-bb01-11ea-8d05-8d135baea311.png)

quotschmacher:
maybe it would be necessary to specify a starting and ending angle and a width, so that the shortage can be shorter?

azivkovic3l:
Thanks for quick reply,

Short term, it would be nice to include this option in examles so everibody
know about it.
Long term, consider shorter length of connection line.

Regards,
Aleksandar

formatc1702:
Yes, I do agree the current implementation is ugly and missing information... any ideas and/or pull requests on this topic are welcome!

quotschmacher:
my python is too weak to help i think.
what about a (overriding) color attribtue for the connection? overriding: override the cable color if a cable is used. and an array in preparation for multi-color wires.

`connections:
  - color: [RD]
    - Relais1: [6, 12, 18 ]
    - Relais1: [12, 18, 24]
`

formatc1702:
Here's a mockup of what it could look like:
![IMG_3700](https://user-images.githubusercontent.com/2447251/87183014-91f0d400-c2e5-11ea-9e12-939116eceefa.JPG)

I'm not sure how easy that would be to represent graphically... I'll think about it.
A possible input syntax could be:
```yaml
connectors:
  X1:
    pinout: [...]
    internal_shorts:
[...]

Issue #45, opened 2020-06-30
[feature] Support for Wago connectors


gioele:
Wago connectors are a common kind of splicing connectors, a modern alternative to twist-on wire connectors.

![wago](https://user-images.githubusercontent.com/237141/86081712-42650980-ba96-11ea-8b5d-7da961bb55d0.jpg)

Differently from in-line splicing connectors (see feature request #26), wires in Wago connectors are all on the same side of the connector. It would be nice if this important property could be carried over in the graphical representation.

aakatz3:
To add onto this, Wire Nuts (common in USA installations) behave relatively similarly, and can likely be implemented at the same time.

![wire-nuts](https://upload.wikimedia.org/wikipedia/commons/0/0b/Junction_box.agr.jpg)

kvid:
The wire nut have all the wires connected in the same "hole" and can probably use the existing ferrule/splice with all wires connected from one side. @formatc1702 should have this in mind when he plan to rename the connector category value.

formatc1702:
Like @kvid mentioned, both wire nuts and Wago 221 series connectors behave very much like a splice. Thus, they should be represented as such in WireViz.

> wires in Wago connectors are all on the same side of the connector. It would be nice if this important property could be carried over in the graphical representation.

For a functional representation of the wiring (which is what WireViz is for), this is irrelevant, and also hard to implement in the current code.

See example 05:
![](https://github.com/formatc1702/WireViz/blob/master/examples/ex05.png?raw=true)
`X1`, `X2` and `X3` are all the same type of connector, but `X2` has wires coming out both sides of each pin due to the daisy chaining.

I will therefore close this issue and refer to #26 for now.

formatc1702:
As a bonus, here's how the image in the first comment might be rendered in WireViz:
(PE rendered as green, since multicolored wires are still in the works)

![test](https://user-images.githubusercontent.com/2447251/86618809-d1ec4a00-bfb9-11ea-82ea-9acdf7051459.png)

I realize the text to generate the connections is not very compact, this is an issue that could be addressed in the future.

```yaml
cables:
  C1: &cable
    type: Mains cable
    colors: [BN, BU, GN] # GN to be replaced with GNYE for PE
    gauge: 1.5
  C2:
    <<: *cable
  C3:
    <<: *cable

connectors:
  S1: &wago
    category: ferrule
    manufacturer: Wago
    manufacturer_part_number: 221-413
    color: OG
  S2:
[...]

PR #75, opened 2020-07-09
Make connecting components together easier and more flexible

### LINKED ISSUE #67: [internal] Make connecting components together easier and more flexible
Inspired by discussions in #26 and #45 .

- Make it easier to define `cable-connector-cable` type connections to make creating [this](https://github.com/formatc1702/WireViz/issues/45#issuecomment-654352415) kind of layouts easier.
  Currently supported connections are `con-cbl-con`, `con-cbl` and `cbl-con`,
  as well as `ferrule-cable` and `cable-ferrule` with ferrule auto-generation.
  Relevant code is [here](https://github.com/formatc1702/WireViz/blob/dev/src/wireviz/wireviz.py#L94L194).
- With some clever coding, maybe even arbitrarily long connection sets could be defined, e.g. `connector-cable-connector-cable-ferrules`
- Allow ferrule-type connectors to have (and show) designators, and include these in the BOM, if visible.
  But probably hide ferrule designators by default.
- Generalize `ferrule` category to `simple`(=only one port/pin, easy to auto-generate if designators are not important)

formatc1702:
Fixes #67.


formatc1702:
Input test file for this branch:
```yaml
connectors:
  X1: &proper
    type: proper connector
    pincount: 4
  F1:
    auto_generate: true
    # category: ferrule
    pincount: 1
    color: PK
  F2:
    auto_generate: true
    # category: ferrule
    pincount: 1
    color: VT
  X2:
    <<: *proper

cables:
  W1:
    wirecount: 4
    color_code: IEC
  W2:
    category: bundle
    wirecount: 4
    color_code: DIN
  W3:
    wirecount: 4
    colors: [RD, YE, GN, BU]
  W99:
    wirecount: 2

connections:
  - # very long connection list, alternating connectors and cables
    - X1: [1-4]        # connector, as usual
    - W1: [2,1,4,3]    # wire, as usual
    - [F1, F1, F1, F2] # list of ferrules, may be different types
    - W2: [1-4]
    - F2               # single ferrule, will be auto-duplicated
    - W3: [2-4,1]
    - X2: [1-4]
  -
    - X1:  [1-2]
    - W99: [1-2]
    - X2:  [1-2]
```

kvid:
Would a connector attribute `hide_single_pin` with default True value help to aboid a lot of special cases for ferrule in the code?

formatc1702:
> Would a connector attribute `hide_single_pin` with default True value help to aboid a lot of special cases for ferrule in the code?

Yes, I am working towards removing ferrules altogether, and implementing something like you suggest.

The new attribute `autogenerate` (no underscore in the future) will only work for connectors with a single pin (ferrules, Wagos with one electrial node only, wire nuts, ...) that don't have to be accessed again in another connection set after being created, and thus can receive internal, auto-incremented identifiers that the user does not need to know.

[This example](https://github.com/formatc1702/WireViz/issues/45#issuecomment-654352415) would _not_ use `autogenerate` for the Wagos since you can't define the three-way splice in one connection set, so you need access to them again. Hoewever, a simplification is in the works here as well, the goal is to make the code work like this:
```yaml
connections:
  -
    - C1: [1-3]
    - [S1, S2, S3] # these have autogenerate set to false, since we need to manually assign names...
    - C3: [1-3]
  -
    - C2: [1-3]
    - [S1, S2, S3] # ...to be able to use them here!
```

src/wireviz/wireviz.py
```diff
@@ -90,108 +58,92 @@ def check_designators(what, where):
...
-            raise Exception('Wrong number of connection parameters')
+    def check_designators(what, where): # helper function
+        for i, x in enumerate(what):
+            if x not in yaml_data[where[i]]:
+                return False
+        return True
+
+    autogenerated_ids = {}
+    for connection in yaml_data['connections']:
+        # TODO: check that items are of alternating type CONNECTOR/FERRULE/FERRULE_LIST and CABLE/WIRE
```
formatc1702, line 70: Ideas?

src/wireviz/wireviz.py
```diff
@@ -90,108 +58,92 @@ def check_designators(what, where):
...
+        # TODO: special case: loops!
```
formatc1702, line 71: Ideas?

src/wireviz/wireviz.py
```diff
@@ -90,108 +58,92 @@ def check_designators(what, where):
...
+                connection_list.append(sublist)
+            elif isinstance(item, dict):  # a component with multiple pins was specified
+                sublist = []
+                id = list(item.keys())[0]
+                pins = expand(list(item.values())[0])
+                for pin in pins:
+                    sublist.append([id, pin])
+                connection_list.append(sublist)
+            elif False:  # TODO: placeholer; a loop inside a connector was specified
+                pass
```
formatc1702, line 123: Ideas?

formatc1702:
@kvid perhaps you have some ideas on how to implement the `TODO`'s I earmarked in the comments above?

Overall I'm quite happy with this refactoring, except for the special case of loops. I'm even considering adding a new section in the YAML file called `loops` or similar, and separating it from `connections`, which should be between different components.

Or maybe even add a `loops` property to the connector itself, since it concerns only one, and potentially a tiny bit of wire. #48 is related but not 100% identical, since I guess some connectors' pins can be shorted internally without requiring a wire loop.

src/wireviz/wireviz.py
```diff
@@ -90,108 +58,92 @@ def check_designators(what, where):
...
+            if itemcount is not None and itemcount_new != itemcount:
+                raise Exception('All lists and dict lists must be the same length!')
+            itemcount = itemcount_new
+        if itemcount is None:
+            raise Exception('No item revealed the number of connections to make!')
+
+        # populate connection list
+        connection_list = []
+        for i, item in enumerate(connection):
+            if isinstance(item, str):  # one single-pin component was specified
```
formatc1702, line 93: It's the responsibility of the user to make sure it's really a single-pin component. If not, the code doesn't care and simply assumes you want to connect pin 1. Same for the next `elif`.

kvid:
I'm afraid I might have to wait until late this evening to test what's stored in the `yaml_data['connections']` in the different cable/wire/connector cases to be able to suggest something, unless you are able to describe briefly what's stored.

One thing about the loops, is that they are often made of short wires, and I would prefer to be able to specify wire attributes for them, and have them in the BOM together with other wires in those cases.

kvid:
Assumimg pin 1 when only a connector without pin is specified, sounds very well in my mind.

formatc1702:
> I'm afraid I might have to wait until late this evening to test what's stored in the `yaml_data['connections']` in the different cable/wire/connector cases to be able to suggest something, unless you are able to describe briefly what's stored.

I've started working on something [here](https://gist.github.com/formatc1702/688fc6047b8744d4c3ed6cafb098d7de), still maturing.

src/wireviz/wireviz.py
kvid: Have you considered this?
`connection_list.append([(id,pin) for pin in pins])`
If the (id,pin) pair is never modified, then I suggest storing it as a tuple instead of a list.

kvid:
```
def conn_item(item):
    if not yaml_data['connectors'][item].get('autogenerate'):
        return item
    autogenerated_ids[item] = autogenerated_ids.get(item, 0) + 1
    new_id = f'_{item}_{autogenerated_ids[item]}'
    harness.add_connector(new_id, **yaml_data['connectors'][item])
    return new_id
```

formatc1702:
Thanks @kvid, I've based the system on your proposal, seems to work great!

src/wireviz/wireviz.py
formatc1702: That won't work. I'll re-implement loops elsewhere, separate from regular connection sets.

formatc1702:
> One thing about the loops, is that they are often made of short wires, and I would prefer to be able to specify wire attributes for them, and have them in the BOM together with other wires in those cases.

This will be its own issue/PR.. for now, I'm just restructuring code, loops will remain as primitive as they have been until now, just easier to declare directly, inside a connector's properties:
```yaml
  X2:
    type: Molex KK 254
    subtype: female
    pinout: [GND, RX, TX, N/C, IN, OUT]
    loops:
      - [5, 6] # each entry must be a list of exactly two pins
      - ...
```

PR #92, opened 2020-07-15
Add type hinting in several places (no change to function)

slightlynybbled:
Made a start on adding standard type hints to functions.  Didn't tackle every function since I'm less familiar with some sections of code than others.


src/wireviz/Harness.py
```diff
@@ -20,16 +20,16 @@ def __init__(self):
...
-    def add_connector(self, name, *args, **kwargs):
+    def add_connector(self, name: str, *args, **kwargs) -> None:
         self.connectors[name] = Connector(name, *args, **kwargs)
 
-    def add_cable(self, name, *args, **kwargs):
+    def add_cable(self, name: str, *args, **kwargs) -> None:
         self.cables[name] = Cable(name, *args, **kwargs)
 
-    def add_bom_item(self, item):
+    def add_bom_item(self, item: str) -> None:
```
formatc1702, line 29: The item is a `dict`, not a `str`.

src/wireviz/Harness.py
```diff
@@ -20,16 +20,16 @@ def __init__(self):
...
         self.additional_bom_items.append(item)
 
-    def connect(self, from_name, from_pin, via_name, via_pin, to_name, to_pin):
+    def connect(self, from_name: str, from_pin: int, via_name: str, via_pin: (int, str), to_name: str, to_pin: int) -> None:
```
formatc1702, line 32: `from_pin` and `to_pin` can be `str` too.

src/wireviz/Harness.py
formatc1702, line 29: The item is a `dict`, not a `str`.

src/wireviz/Harness.py
formatc1702, line 32: `from_pin` and `to_pin` can be `str` as well.

src/wireviz/Harness.py
slightlynybbled: Thanks.  I haven't used this directly yet.  Fixed in most recent push.

src/wireviz/Harness.py
slightlynybbled: Fixed in most recent push

Issue #63, opened 2020-07-05
Include a set of input files for regression testing


kvid:
During the development, a lot of test input files have been created to test a lot of corner cases. I suggest including most of these together with the direct output files to enable regression testing. The output files created by Graphviz might differ with different Graphviz versions, and is therefore not very suited for regression testing, but the .gv and .tsv files could be checked by 'git diff ../../{examples,tutorial,corner_cases}/*.tsv' as part of the PR checks.

"I have tested it with the files I used for testing the part numbers and they all generated fine."

_Originally posted by @Tyler-Ward in https://github.com/formatc1702/WireViz/pull/59#issuecomment-653756012_

Tyler-Ward:
This sounds like a good idea, could also use pytest to compare the generated data against the pre existing data without needing extra commands. Would need access to the gv and tsv data in a similar way to the svg in #55 in order to do this however.

kvid:
My argument relared to this issue in https://github.com/formatc1702/WireViz/pull/94#issuecomment-659995911
> Personally, I'm actively using the input files together with the output files (those that don't depend on the Graphviz version) for regression testing to verify that my commits either don't change any output or produce the expected output diff. That's also why I suggested issue #63
> 
> I would therefore prefer to keep these files unless the regression testing can be secured some other way. Why not change the clean action to do git checkout -- {generated files} or something similar?

kvid:
Responce from @formatc1702 in https://github.com/formatc1702/WireViz/issues/100#issuecomment-660604620
> Regarding the output files:
> 
> I agree doing diffs for regression testing is a good idea.
> 
> Probably the biggest annoyance is the differing GraphViz version causing unneeded diffs.
> 
> One option is to have the clean option delete the .gv files, since they are of little interest after development of a feature is finished.
> Regarding the use of a _built directory or similar: I don't understand the first of your points:
> 
> Two advantages by copying the YAML file to the subfolder first:
> 
> - Avoid changes to the Harness class.
> - Avoid bad references from the readme files.
> 
> What do you mean by avoiding changes to the Harness class?

kvid:
The YAML input below test all color codes currently defined in `_color_hex` and `COLOR_CODES`. Until bugfix #145 is accepted, the error message `Unknown color specified` is repeated 40 times.
```yaml
# Generated by build_colors.py
connectors:
  T568B:
    show_pincount: false
    pinlabels: [white/orange, orange, white/green, blue, white/blue, green, white/brown, brown]
[...]

PR #94, opened 2020-07-16
Allow cleaning of examples

aakatz3:
build_examples supports cleaning examples and intelligently will detect new examples.

SUGGESTION: When merging into dev, require
```
build_examples.py clean
```

and then only build when merging into master branch

This should allow easier merges, especially from windows, and I am using it as part of a resolution to one of the requests in #17

### DIFF
--- a/src/wireviz/build_examples.py
+++ b/src/wireviz/build_examples.py
@@ -1,57 +1,110 @@
 #!/usr/bin/python3
 # -*- coding: utf-8 -*-
-
+import argparse
 import os
 import sys
+from fnmatch import fnmatch
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
 
 from wireviz import wireviz
 
-demos     = 2 # 2
-examples  = 9 # 9
-tutorials = 8 # 7
+examples_path = os.path.join('..','..','examples')
+tutorials_path = os.path.join('..','..','tutorial')
+demos_path = examples_path
 
-if demos:
-    for i in range(1,demos+1):
-        fn = '../../examples/demo{:02d}.yml'.format(i)
-        print(fn)
-        wireviz.parse_file(fn)
+readme = 'readme.md'
 
-if examples:
-    with open(os.path.abspath('../../examples/readme.md'), 'w') as file:
+
+def build_demos():
+    for fn in os.listdir(demos_path):
+        if fnmatch(fn, "demo*.yml"):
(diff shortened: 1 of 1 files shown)


formatc1702:
This looks like a really useful improvement! Thanks!

formatc1702:
Any idea how it chooses the order of the files? Any quick way to fix this?
See my console output when I run the script:

```
(venv) wireviz $ python build_examples.py 
../../examples/ex09.yml
../../examples/ex08.yml
../../examples/ex01.yml
../../examples/ex03.yml
../../examples/ex02.yml
../../examples/ex06.yml
../../examples/ex07.yml
../../examples/ex05.yml
../../examples/ex04.yml
../../examples/demo01.yml
../../examples/demo02.yml
../../tutorial/tutorial08.yml
../../tutorial/tutorial02.yml
../../tutorial/tutorial03.yml
../../tutorial/tutorial01.yml
../../tutorial/tutorial04.yml
../../tutorial/tutorial05.yml
../../tutorial/tutorial07.yml
../../tutorial/tutorial06.yml
```

formatc1702:
Also, I'm not sure cleaning up the examples is the best course of action, because then the [readme of the `dev` branch](https://github.com/formatc1702/WireViz/tree/dev) would include references to nonexistent files; same for the tutorial and example galleries.

What I've been doing is simply discarding any changes I've made to the output files before commiting anything to `dev`.

As the project matures, there probably will be less and less changes that affect the output files, but who knows.

formatc1702:
Perhaps one option would be to keep examples out of the repo completely (by `.gitignore`ing the entire directory), and curating a separate `wireviz-gallery` repo that can be rebuilt and updated at will. The WireViz readme could directly link to those files on the other repo. This would avoid links to missing files.

kvid:
Personally, I have actively used the input files together with the output files that don't depend on the Graphviz version as regression tests to verify that my commits either don't change any output or produce the expected output diff. That's also why I suggested issue https://github.com/formatc1702/WireViz/issues/63

I would therefore prefer to keep these files unless the regression testing can be secured some other way. Why not change the `clean` action to do `git checkout -- {generated files}` or something similar?

kvid:
If `build_examples.py` is changed to write generated files to a `_built` subfolder unless some action argument specify to overwrite the committed files, then it will be easy both to compare differences side by side and to ignore the subfolder when committing. What do you think?

Issue #88, opened 2020-07-14
[internal] Match order of parse() return tuple items to order of input tuple


formatc1702:
Currently, if the `parse()` function of the `wireviz` module is run with, e.g. `return_types = ('svg', 'png')`, the returned tuple will still include the PNG before the SVG.

@slightlynybbled perhaps you could propose a fix, so that the order of the return tuple items matches the ones requested in `return_types`?

formatc1702:
Well, that was easy :) I guess I was just too tired yesterday to come up with a quick solution. Thanks!

Issue #86, opened 2020-07-14
[bug] \n Displayed in Cable Notes


davidrea:
Adding a `\n` escaped character works to insert a line break in:

* Connector `notes:` entities
* Connector `type:` and `subtype:` entities

But the `\n` escaped character is printed in cable `notes:` entities:

![image](https://user-images.githubusercontent.com/229141/87438489-97883b80-c5bd-11ea-93e7-d96805ce0ba1.png)

Source for above:

```yml
connectors:

  P1:
    type: Socket housing
    subtype: Molex SL\n50579406
    pinout: [VDD, SWCLK, GND, SWDIO, NRST, SWO]
    notes: "Terminal: 16020086 or equiv."

  P2:
    type: Pin header
    subtype: Adam Tech\nBHR-20-VUA\nor Equiv.
    pinout: [VCC, VCC (optional), TRST, GND, TDI, GND, DIO/TMS, GND, DCK/TCK, GND, RTCK, GND, SWO/TDO, GND, NRST, GND, NC, GND, NC, GND]
    notes: Solder wires to pins,\nheat shrink individually

cables:
  W1:
    gauge: 24 AWG
    length: 0.2
    wirecount: 6
    category: bundle
    colors: [RD, GY, BK, GY, GY, GY]
    notes: Stranded,\nHeat shrink bundle 2 plcs
    # ^^^ this line contains printed \n ^^^

connections:
  -
    - P1: [1, 2, 3, 4, 5, 6]
    - W1: [1, 2, 3, 4, 5, 6]
    - P2: [1, 9, 4, 7, 15, 13]
```

davidrea:
Just tacking onto this, \n escaped characters also appear in HTML BOM:

![image](https://user-images.githubusercontent.com/229141/87444466-d53c9280-c5c4-11ea-9fde-8dd7d924ce7b.png)

formatc1702:
Thanks for the heads up. I can't look at the issue in detail now, but I can offer solution that (IMHO) is more elegant than using `\n` for new lines, since [YAML supports multiline strings natively](https://yaml-multiline.info/).
The code below fixes both the rendering and BOM issues:

```yaml
connectors:
  P1:
    type: Socket housing
    subtype: |
      Molex SL
      50579406
    pinout: [VDD, SWCLK, GND, SWDIO, NRST, SWO]
    notes: "Terminal: 16020086 or equiv."

  P2:
    type: Pin header
    subtype: |
      Adam Tech
[...]

PR #95, opened 2020-07-16
Fix BOM portion of generated HTML files

slightlynybbled:


slightlynybbled:
Also fixes BOM display as noted in #86.

src/wireviz/Harness.py
```diff
@@ -321,8 +321,10 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
             for row in listy[1:]:
                 file.write('<tr>')
                 for i, item in enumerate(row):
+                    item_str = item.replace('\u00b2', '<sup>2</sup>')
```
formatc1702, line 324: I wonder if using [`&sup2;`](https://www.toptal.com/designers/htmlarrows/punctuation/superscript-two/) will work better in this case, since it should preserve the correct character when copying & pasting from the HTML file?

slightlynybbled: Your call.  I'm not an HTML guy, I just knew how to fix it.

formatc1702: Yup, [just tried it](https://www.w3schools.com/code/tryit.asp?filename=GGTWHOIGBPI6); please use `&sup2;` :)
Currently there's a conflict, but for such a small change, it should be easy to resolve.
Thanks!

slightlynybbled:
The extra slash is an escape character in python strings. Actual string being checked  is ‘\n’.

kvid:
> The extra slash is an escape character in python strings. Two slashes are required to represent one slash. Actual string being checked is ‘\n’

I know that. :-) When I write `'\\n'` in code formatting, I mean a two-character string: backslash followed by the letter n. The user might put that in the input text as an attempt to obtain a line break. Currently, this only works for connector attributes in the diagram, and with your commit, also in the HTML BOM, but all other output show it literally as backslash followed by the letter n.

As @formatc1702 describes in his issue #86 recommendation, using YAML multiline strings is a better approach to obtain line breaks, that insert actual newline characters, and that converting `'\\n'` to newline is then unwanted:
> Looking at the Flow Scalars section on yaml-multiline.info, it seems that not converting \n into a newline for plain strings is the expected behavior.

Actual newline characters in the input text is currently supported for all output (either converted to a space or to a proper line break for each output).

formatc1702:
> If we should not support `'\\n'` in the input text, then we let the user easily discover his mistake, not hide it in some output.

I agree with @kvid on this one; if the YAML parser does not generate newlines with `\n`, why should we?
As described in #88, if (for whatever reason) the user wants to use `\n`, it needs to be within a double quoted string and thus, again, the YAML parser itself will take care of it.

No need for an extra check here!

slightlynybbled:
I will make the change b/c this is not my package, but I disagree.

 - The `\n` character has been a new line for most of modern computer history.  It is "the way" to create a line break and, if it can be easily supported, it should be.
 - Many/Most users are only using YAML b/c it is the author's choice and are likely not as familiar with YAML syntax.
 - There is currently no documentation for this package to guide the user, so the package should be forgiving.

formatc1702:
Thanks.

> I will make the change b/c this is not my package, but I disagree.
>
>   - The `\n` character has been a new line for most of modern computer history.  It is "the way" to create a line break and, if it can be easily supported, it should be.

It can be easily supported by adding the double quotes.

>  - Many/Most users are only using YAML b/c it is the author's choice and are likely not as familiar with YAML syntax.
>  - There is currently no documentation for this package to guide the user, so the package should be forgiving.

I agree on the last point. IMHO, this is an argument for better documentation (which we absolutely need to do), but not to deviate from the clearly defined YAML syntax.

It may seem like a far-fetched corner case, but an example that comes to mind is somebody using a Windows path in the notes field `C:\Documents\new harness\file.ext` and wondering why it's split.

kvid:
I also respect your opinion, and I fully agree that not everyone has full knowledge about YAML syntax (including myself). Please create a new issue where you suggest how to guide the user about line breaks and similar issues. Maybe some additions to README, tutorials, and/or examples? You'll find a lot of examples on [yaml-multiline.info](https://yaml-multiline.info/) - please create a few examples for this project based on what you think the users might need, e.g. something like this:
```
connectors:
  X1:
    pincount: 4
    notes: |
        First line
        Second line
  X2:
    pincount: 4
    notes: "First line\nSecond line"
```

PR #102, opened 2020-07-19
Unify connectors

### LINKED ISSUE #78: [internal] Merge ferrules with connectors
- Code will be simplified, and a separate `ferrule` category will no longer be needed.
- Connectors with one pin can optionally be rendered in a simplified way, with no pin table, by using a `style: simple` (or similar, to be defined) attribute in the YAML input.
- Connectors with one pin can optionally be auto-generated (with auto-assigned identifiers that will not be shown in the BOM) by using the `autogenerate: true` attribute in the YAML input.

formatc1702:
Closes #78.
- Code has been adapted, all references to "ferrules" as separate connector type have been eliminated.
- The simplified look of a ferrule can be achieved by adding a `style: simple` attribute to a connector.
  - This simply eliminates the list of pins inside the node, and causes connected wires to attach to the node itself, rather than a subport
  - The `simple` style is only allowed for connectors with one [electrical] port:
    - ferrules
    - splices (may be two-way, three-way or more)
    - Wagos / Wire nuts (see #45)
- The auto-generation of simple connectors is achieved by the `autogenerate: true` attribute
- Autogenerated connectors hide their (also autogenerated) name by default, and their designators are not listed in the respective BOM column
- Some examples have been updated/simplified to work with these changes


src/wireviz/DataClasses.py
```diff
@@ -55,6 +59,18 @@ def __post_init__(self):
         if len(self.pinnumbers) != len(set(self.pinnumbers)):
             raise Exception('Pin numbers are not unique')
 
+        if self.show_name is None:
+            if self.autogenerate:
+                self.show_name = False  # hide auto-generated designators by default
+            else:
+                self.show_name = True
```
kvid, line 66: The if-else spanning lines 63-66 might be replaced by:
`self.show_name = not self.autogenerate  # hide auto-generated designators by default`

src/wireviz/DataClasses.py
```diff
@@ -55,6 +59,18 @@ def __post_init__(self):
...
+
+        if self.show_pincount is None:
+            if self.style == 'simple':
+                self.show_pincount = False  # hide pincount for simple connectors, since they are 1 pin connectors by definition
+            else:
+                self.show_pincount = True
```
kvid, line 72: The if-else spanning lines 69-72 might be replaced by:
`self.show_pincount = self.style != 'simple'  # hide pincount for simple (1 pin) connectors by default`

src/wireviz/DataClasses.py
formatc1702: Done.

src/wireviz/DataClasses.py
formatc1702: Done.

kvid:
When I run `python build_examples.py`, I get some unexpected links in the readme.md files, like this:
```
![](tutorial../../tutorial/tutorial01.png)

[Bill of Materials](tutorial../../tutorial/tutorial01.bom.tsv)
```
Am I doing something wrong, or might this be caused by an earlier merge-in to dev?

PR #108, opened 2020-07-20
Refactor build_examples.py

formatc1702:
- Use `pathlib.Path` instead of `os.path`
- Fix order of files while building
- Consolidate code for building demos, examples and tutorial
- Change argument from `tutorials` to `tutorial` to remain consistent
- Add some indentation in console output for better readability


formatc1702:
Not sure what happened here. New PR #110 is open.

PR #110, opened 2020-07-20
Refactor build_examples.py

formatc1702:
- Use `pathlib.Path` instead of `os.path`
- Fix order of files while building
- Consolidate code for building demos, examples and tutorial
- Change argument from `tutorials` to `tutorial` to remain consistent
- Add some indentation in console output for better readability


kvid:
I would like to include a new action to restore generated files using `git checkout -- {all generated files}`. What do you think? It does _not_ have to replace the current `clean` action as I suggested in my argument copied into https://github.com/formatc1702/WireViz/issues/63#issuecomment-660673948.

formatc1702:
> I would like to include a new action to restore generated files using `git checkout -- {all generated files}` or something similar. What do you think? It does _not_ have to replace the current `clean` action as I suggested in my argument copied into [#63 (comment)](https://github.com/formatc1702/WireViz/issues/63#issuecomment-660673948).

Sounds like a good idea; I would welcome a code contribution for this!

kvid:
Does this PR address this issue? https://github.com/formatc1702/WireViz/commit/0da444167889d9af7d77616bf2d977ad308100f7#r40750948

formatc1702:
Yes, see latest commit :) Thanks

aakatz3:
Why remove the upload action? I don't see how the build results can be verified without it, unless there is a new action that performs regression testing. Otherwise, the outputs are generated and ignored. It shouldn't count against any git quotas, I don't see a reason to remove it, but please let me know if I'm missing something.


Great work on the refactor!

kvid:
@aakatz3 Can you briefly explain what the upload action does or link to a description. I want to learn. :-)

aakatz3:
Sure! [This is the action it uses](https://github.com/actions/upload-artifact). The way it is configured, it essentially zips up the examples and tutorial folders after github actions runs the build test, and uploads them so that a user can download them, to manually verify them. They appear here: 
image

Clicking the link will provide you with a .zip file, which has the directories that are specified in the action to upload.

image

Inside those folders are all of the files that are in the folders after running all previous tasks in the github action, I.E. installing graphviz and dependancies, and generating the examples. 
image

This is a particularly useful tool because, while an individual developer's environment may differ, the VM that builds the github action is always the same, which means it is perfect for both automated and manual regression testing. You could either then take the uploaded artifact and pass it to some external testing service, or download it and manually examine the generated images and other outputs, to make sure they match the ones you generate on your system, and work as expected. 


If you would like more information, I can try to find a slightly easier tutorial than the repo for the action itself, as it is slightly confusing.

formatc1702:
Improved refactoring is being taken care of in #118, closing.

PR #113, opened 2020-07-20
Change code for silver and gold

formatc1702:
Change to SR and GD to match the [standard](https://en.wikipedia.org/wiki/Electronic_color_code#Color_band_system)


aakatz3:
Makes sense, I just used a swatch I had on hand, completely forgetting there are standards for that. To be honest, I just kinda used Solidworks.

Issue #50, opened 2020-07-01
[feature] Ability to add BOM items within components


MikeCrookes:
Would be great to be able to specify the crimps for connectors and for this to be added to the BOM

formatc1702:
This is a very good point, especially once you get to more complicated [mixed contact connectors](https://d-sub-connector.com/pages/combi-d-sub-connectors) (first example that come to mind) which need different crimp types within one connector.

Maybe @Tyler-Ward our BOM expert has a suggestion? 😃

Tyler-Ward:
This sounds like a good idea and probably wouldn't be too hard to add. Doing #51 first sounds sensible and should be relatively easy to do.

I can see there being two methods to define the crimps.

For single crimp types a string can be set for the crimp part number and internal part number and used for all positions. (might want a flag to turn on/off limiting crimps to only being added for populated positions). With Templating this can allow the crimp to be declared for an entire series of connectors at once e.g. all female Molex KKs.

For multiple crimp types we can specify with arrays similar to bundle wires. Will likely also need a field for a crimp description so they can have a useful description in the bom. The array helper function in #62 will help for this. As well as the example above a common example of needing this might be where there are different crimp parts depending on wire gauge.

For interest If you want connectors with unusual loadouts I believe it is possible to set up some lemo connectors to have all of the following in one connector, Low voltage, high voltage, coax, fibre-optic and fluid/gas.

formatc1702:
Do you want to take care of this? I'm starting to assign issues to contributors to keep an overview of what's going on, so let me know!

Tyler-Ward:
I'm happy to take those on so feel free to assign to me.

formatc1702:
Much appreciated. It seems you have to comment on #51 in order to be eligible for assigning.

formatc1702:
An idea here would be to tag certain items to be used only for pins that are actually populated, others for all pins (qty = pincount), others in custom quantities (like you mentioned, backshells and the like)

Tyler-Ward:
Apologies for the lack of progress on this the last couple of weeks have bee quite busy. rather than the implementation idea suggested above in https://github.com/formatc1702/WireViz/issues/50#issuecomment-653888954 integrating crimps into the idea from #51 which supports additional components on parts might be a better soloution rather than having two methods of defining.

[...]

Issue #114, opened 2020-07-20
[internal] Proposal: shorten BOM part number attributes


formatc1702:
I propose shortening the part number attributes from
```yaml
    manufacturer: ABC
    manufacturer_part_number: 123
    internal_part_number: 456
```
to
```yaml
    pn: 456
    mpn: 123
    manufacturer: ABC
```

MPN seems to be an industry standard abbreviation.
Also, they are [currently rendered](https://github.com/formatc1702/WireViz/blob/dev/tutorial/tutorial08.png) as `MPN` and `IPN` anyway.

I propose `pn` over `ipn` in the YAML, and rendering it as `P/N: ` in the output, but that's up for debate.

@Tyler-Ward (and others, of course) any thoughts?

[Edit]
As a side note, I would also like to swap the order of the fields in the graphic output to match the above:
`P/N: 456 | MPN: 123 | ABC`

Tyler-Ward:
This seems sensible to me. I usualy go for longer names to avoid confustion but these look just as clear and take up less space.

Swapping to P/N rather than ipn sounds good, removing the explicit internal could also allow for people to use that field for a distributor/supplier part number if they dont have an internal part numbering system.

The order swap sounds logical as the pn field will likley contain the most usefull information in that group.

aakatz3:
I agree, only suggestion is for MPN to render as MP/N or M P/N.

formatc1702:
Perhaps the most elegant option would be to show only two fields in the node:
`P/N: {pn} | {manufacturer}: {mpn}`

And replace `{manufacturer}` with `MPN`/`MP/N`/some variant only if the field is `None`.
My thinking is that it will be rare to supply an MPN without also specifying a manufacturer... and even if the manufacturer appears e.g. in the `type` attribute (like the many `Molex KK 254` in the examples), you'd probably still want it in a separate BOM field, for sorting/filtering by manufacturer easily.

> removing the explicit internal could also allow for people to use that field for a distributor/supplier part number if they dont have an internal part numbering system.

My thinking almost goes the other way... people could use `manufacturer`+`mpn` to specify e.g. `Digikey: 12345678` which would be more practical, for the hobbyist level.

Obviously you could go wild, allowing lists in the fields to specify multiple alternative suppliers and a number for each, but IMHO if you (as a user) are that serious, you should be looking at an ERP and internal numbering systems, and keep track of supplier variants there. In other words, do NOT think I'm advocating for adding even more fields/flexibility here :)

[...]

PR #111, opened 2020-07-20
Syntax documentation + new /docs directory

formatc1702:
Closes #107.

See current status in rendered form [here](https://github.com/formatc1702/WireViz/blob/feature/syntax-description/syntax.md).

Issue #101, opened 2020-07-19
[meta] Collaborator discussion


formatc1702:

formatc1702:
I've created [a new milestone](https://github.com/formatc1702/WireViz/milestone/1) to show which features I'd like to have included in the next release, i.e. pushing to `master` as well as uploading the new package to [PyPI](https://pypi.org/project/wireviz/).

If anybody wants to suggest more issues to include, please let me know. However, my aim is to release v0.2 relatively soon, which is why I've deliberately left out a lot of existing requests.

formatc1702:
Uptate on the [v0.2 milestone](https://github.com/formatc1702/WireViz/milestone/1):
I've assigned people for most open issues/PRs, hoping to be able to assign #114 soon too. If anybody wants to take on #60, please comment there and I will asign you.

Additionally, I've made a [milestone for v0.3](https://github.com/formatc1702/WireViz/milestone/2), which currently only serves to denote issues/PR drafts that I know I want to pursue as soon as v0.2 is out, but not before.

kvid:
Have you considered #116 and #121 for the [v0.2 milestone](https://github.com/formatc1702/WireViz/milestone/1)?

formatc1702:
Added both! Thanks.

kvid:
I just made a suggestion in the closed issue https://github.com/formatc1702/WireViz/issues/122#issuecomment-667545193. Do you think it should be reopened?

formatc1702:
Feel free to do so!

kvid:
> Feel free to do so!

I don't know how. Maybe I'm not authorized to do it.

formatc1702:
Not sure.. did it myself now :)

kvid:
> Not sure.. did it myself now :)

Thank's

> 
> Perhaps only repo owners (and maybe the original commenter) have the "reopen issue" button...

You are right. I certainly did not see that button here, but I can see it in one of the closed issues I opened myself.

formatc1702:
Just a quick FYI especially for @kvid, I'll be taking a couple of days off coding to breathe a bit, and I promise I shall be back to read through all the comments, commits and other contributions as usual.
It will help me get back my focus and ensure v0.2 gets released with no major issues :)

formatc1702:
My goal is to get [v0.2](https://github.com/formatc1702/WireViz/milestone/1) released by the end of the week.
- #27 / #153 is almost ready for merging
[...]

PR #115, opened 2020-07-20
Feature/additional components

Tyler-Ward:
Draft implementation of #50.

An exmple file to work with this is shown below.

```yaml
connectors:
  X1:
    type: Powerpole
    pinout: [GND, VCC]
    ignore_in_bom: true
    additional_components:
      -
        type: housing
        qty: 1
        unit: ea
        manufacturer: Anderson
        manufacturer_part_number: 1327G6
        internal_part_number: powerpole-black
      -
        type: housing
        qty: 1
        unit: ea
        manufacturer: Anderson
        manufacturer_part_number: 1327
        internal_part_number: powerpole-red
      -
        type: crimp
        qty: pincount
        unit: ea
        manufacturer: Anderson
        manufacturer_part_number: 1331
        internal_part_number: powerpole-crimp-30A
  X2: &mf3-5
    type: Molex Micro-Fit
    subtype: female
    pinout: [GND, VCC, A, B, C]
    manufacturer_part_number: 436450500
    manufacturer: Molex
    internal_part_number: MF3.0-5
    additional_components:
      -
        type: crimp
        qty: connectioncount
        unit: ea
        manufacturer: Molex
        manufacturer_part_number: 43030-0007
        internal_part_number: MF3.0-Crimp

  X3:
    <<: *mf3-5

cables:
    show_equiv: true
    length: 0.2
    manufacturer_part_number: 123456

  W2:
    <<: *wire_power

connections:
  -
    - X1: [1-2]
    - W1: [1-2]
    - X2: [1-2]
  -
    - X1: [1-2]
    - W2: [1-2]
    - X3: [1-2]
```

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from wireviz.wv_helper import int2tuple, aspect_ratio
 from wireviz import wv_colors
 
+# Literal type aliases below are commented to avoid requiring python 3.8
 ConnectorMultiplier = str  # = Literal['pincount', 'populated']
 CableMultiplier = str  # = Literal['wirecount', 'terminations', 'length', 'total_length']
 
@@ -59,7 +60,7 @@ class AdditionalComponent:
 
     @property
     def description(self) -> str:
-        return self.type.capitalize().strip() + (f', {self.subtype.strip()}' if self.subtype else '')
+        return self.type.rstrip() + (f', {self.subtype.rstrip()}' if self.subtype else '')
 
 
 @dataclass
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -59,8 +59,7 @@ class AdditionalComponent:
 
     @property
     def description(self) -> str:
-        name_subtype = f', {self.subtype}' if self.subtype else ''
-        return f'{self.type.capitalize()}{name_subtype}'
+        return self.type.capitalize() + (f', {self.subtype}' if self.subtype else '')
 
 
 @dataclass
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -66,6 +66,7 @@ class Connector:
     autogenerate: bool = False
     loops: List[Any] = field(default_factory=list)
     ignore_in_bom: bool = False
+    additional_components: List[Any] = None
 
     def __post_init__(self):
 
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -43,6 +43,21 @@ class Image:
                 if self.width:
                     self.height = self.width / aspect_ratio(gv_dir.joinpath(self.src))
 
+@dataclass
+class AdditionalComponent:
+    type: str
+    subtype: Optional[str] = None
+    manufacturer: Optional[str] = None
+    mpn: Optional[str] = None
+    pn: Optional[str] = None
+    qty: float = 1
+    unit: Optional[str] = None
+    qty_multiplier: Optional[str] = None
+
+    def long_name(self) -> str:
+        name_subtype = f', {self.subtype}' if self.subtype else ''
+        return f'{self.type.capitalize()}{name_subtype}'
+
 
 @dataclass
 class Connector:
@@ -66,7 +81,7 @@ class Connector:
     autogenerate: bool = False
     loops: List[Any] = field(default_factory=list)
     ignore_in_bom: bool = False
-    additional_components: List[Any] = None
+    additional_components: List[AdditionalComponent] = field(default_factory=list)
 
     def __post_init__(self):
 
@@ -116,10 +131,9 @@ class Connector:
             if len(loop) != 2:
                 raise Exception('Loops must be between exactly two pins!')
 
-        if self.additional_components:
-            for additional_component in self.additional_components:
-                if 'type' not in additional_component:
-                    raise Exception('Additional components must have a type specified')
+        for i, item in enumerate(self.additional_components):
+            if isinstance(item, dict):
+                self.additional_components[i] = AdditionalComponent(**item)
 
     def activate_pin(self, pin):
         self.visible_pins[pin] = True
@@ -147,7 +161,7 @@ class Cable:
     show_name: bool = True
     show_wirecount: bool = True
     ignore_in_bom: bool = False
-    additional_components: List[Any] = None
+    additional_components: List[AdditionalComponent] = field(default_factory=list)
 
     def __post_init__(self):
 
@@ -205,11 +219,9 @@ class Cable:
                 else:
                     raise Exception('lists of part data are only supported for bundles')
 
-        if self.additional_components:
-            for additional_component in self.additional_components:
-                if 'type' not in additional_component:
-                    raise Exception('Additional components must have a type specified')
-
+        for i, item in enumerate(self.additional_components):
+            if isinstance(item, dict):
+                self.additional_components[i] = AdditionalComponent(**item)
 
     def connect(self, from_name, from_pin, via_pin, to_name, to_pin):
         from_pin = int2tuple(from_pin)
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -10,6 +10,7 @@ from wireviz import wv_colors
 ConnectorMultiplier = str  # = Literal['pincount', 'populated']
 CableMultiplier = str  # = Literal['wirecount', 'terminations', 'length', 'total_length']
 
+
 @dataclass
 class Image:
     gv_dir: InitVar[Path] # Directory of .gv file injected as context during parsing
@@ -56,7 +57,7 @@ class AdditionalComponent:
     unit: Optional[str] = None
     qty_multiplier: Union[ConnectorMultiplier, CableMultiplier, None] = None
 
-    def long_name(self) -> str:
+    def description(self) -> str:
         name_subtype = f', {self.subtype}' if self.subtype else ''
         return f'{self.type.capitalize()}{name_subtype}'
 
@@ -139,6 +140,16 @@ class Connector:
 
     def activate_pin(self, pin):
         self.visible_pins[pin] = True
+
+    def get_qty_multiplier(self, qty_multiplier: Optional[ConnectorMultiplier]) -> int:
+        if not qty_multiplier:
+            return 1
+        elif qty_multiplier == 'pincount':
+            return self.pincount
+        elif qty_multiplier == 'populated':
+            return sum(self.visible_pins.values())
+        else:
+            raise ValueError(f'invalid qty multiplier parameter for connector {qty_multiplier}')
 
 
 @dataclass
@@ -234,6 +245,20 @@ class Cable:
         for i, _ in enumerate(from_pin):
             # self.connections.append((from_name, from_pin[i], via_pin[i], to_name, to_pin[i]))
             self.connections.append(Connection(from_name, from_pin[i], via_pin[i], to_name, to_pin[i]))
+
+    def get_qty_multiplier(self, qty_multiplier: Optional[CableMultiplier]) -> float:
+        if not qty_multiplier:
+            return 1
+        elif qty_multiplier == 'wirecount':
+            return self.wirecount
+        elif qty_multiplier == 'terminations':
+            return len(self.connections)
+        elif qty_multiplier == 'length':
+            return self.length
+        elif qty_multiplier == 'total_length':
+            return self.length * self.wirecount
+        else:
+            raise ValueError(f'invalid qty multiplier parameter for cable {qty_multiplier}')
 
 
 @dataclass
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -142,6 +142,7 @@ class Cable:
     show_name: bool = True
     show_wirecount: bool = True
     ignore_in_bom: bool = False
+    additional_components: List[Any] = None
 
     def __post_init__(self):
 
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -7,6 +7,8 @@ from pathlib import Path
 from wireviz.wv_helper import int2tuple, aspect_ratio
 from wireviz import wv_colors
 
+ConnectorMultiplier = str  # = Literal['pincount', 'populated']
+CableMultiplier = str  # = Literal['wirecount', 'terminations', 'length', 'total_length']
 
 @dataclass
 class Image:
@@ -52,7 +54,7 @@ class AdditionalComponent:
     pn: Optional[str] = None
     qty: float = 1
     unit: Optional[str] = None
-    qty_multiplier: Optional[str] = None
+    qty_multiplier: Union[ConnectorMultiplier, CableMultiplier, None] = None
 
     def long_name(self) -> str:
         name_subtype = f', {self.subtype}' if self.subtype else ''
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -116,6 +116,10 @@ class Connector:
             if len(loop) != 2:
                 raise Exception('Loops must be between exactly two pins!')
 
+        for additional_component in self.additional_components:
+            if 'type' not in additional_component:
+                raise Exception('Additional components must have a type specified')
+
     def activate_pin(self, pin):
         self.visible_pins[pin] = True
 
@@ -199,6 +203,10 @@ class Cable:
                         raise Exception('lists of part data must match wirecount')
                 else:
                     raise Exception('lists of part data are only supported for bundles')
+
+        for additional_component in self.additional_components:
+            if 'type' not in additional_component:
+                raise Exception('Additional components must have a type specified')
 
 
     def connect(self, from_name, from_pin, via_pin, to_name, to_pin):
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -57,6 +57,7 @@ class AdditionalComponent:
     unit: Optional[str] = None
     qty_multiplier: Union[ConnectorMultiplier, CableMultiplier, None] = None
 
+    @property
     def description(self) -> str:
         name_subtype = f', {self.subtype}' if self.subtype else ''
         return f'{self.type.capitalize()}{name_subtype}'
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -65,6 +65,7 @@ class Connector:
     hide_disconnected_pins: bool = False
     autogenerate: bool = False
     loops: List[Any] = field(default_factory=list)
+    ignore_in_bom: bool = False
 
     def __post_init__(self):
 
@@ -139,6 +140,7 @@ class Cable:
     color_code: Optional[str] = None
     show_name: bool = True
     show_wirecount: bool = True
+    ignore_in_bom: bool = False
 
     def __post_init__(self):
 
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -59,7 +59,7 @@ class AdditionalComponent:
 
     @property
     def description(self) -> str:
-        return self.type.capitalize() + (f', {self.subtype}' if self.subtype else '')
+        return self.type.capitalize().strip() + (f', {self.subtype.strip()}' if self.subtype else '')
 
 
 @dataclass
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -116,9 +116,10 @@ class Connector:
             if len(loop) != 2:
                 raise Exception('Loops must be between exactly two pins!')
 
-        for additional_component in self.additional_components:
-            if 'type' not in additional_component:
-                raise Exception('Additional components must have a type specified')
+        if self.additional_components:
+            for additional_component in self.additional_components:
+                if 'type' not in additional_component:
+                    raise Exception('Additional components must have a type specified')
 
     def activate_pin(self, pin):
         self.visible_pins[pin] = True
@@ -204,9 +205,10 @@ class Cable:
                 else:
                     raise Exception('lists of part data are only supported for bundles')
 
-        for additional_component in self.additional_components:
-            if 'type' not in additional_component:
-                raise Exception('Additional components must have a type specified')
+        if self.additional_components:
+            for additional_component in self.additional_components:
+                if 'type' not in additional_component:
+                    raise Exception('Additional components must have a type specified')
 
 
     def connect(self, from_name, from_pin, via_pin, to_name, to_pin):
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -378,7 +378,7 @@ class Harness:
                                + (f', {connector.pincount} pins' if connector.show_pincount else '')
                                + (f', {connector.color}' if connector.color else ''))
                 bom_entries.append({
-                    'item': description, 'qty': 1, 'unit': '', 'designators': connector.name if connector.show_name else None,
+                    'item': description, 'qty': 1, 'unit': None, 'designators': connector.name if connector.show_name else None,
                     'manufacturer': connector.manufacturer, 'mpn': connector.mpn, 'pn': connector.pn
                 })
 
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -202,7 +202,7 @@ class Harness:
                 wirehtml.append('     </table>')
                 wirehtml.append('    </td>')
                 wirehtml.append('   </tr>')
-                if(cable.category == 'bundle'):  # for bundles individual wires can have part information
+                if cable.category == 'bundle':  # for bundles individual wires can have part information
                     # create a list of wire parameters
                     wireidentification = []
                     if isinstance(cable.pn, list):
@@ -213,7 +213,7 @@ class Harness:
                     if manufacturer_info:
                         wireidentification.append(html_line_breaks(manufacturer_info))
                     # print parameters into a table row under the wire
-                    if(len(wireidentification) > 0):
+                    if len(wireidentification) > 0 :
                         wirehtml.append('   <tr><td colspan="3">')
                         wirehtml.append('    <table border="0" cellspacing="0" cellborder="0"><tr>')
                         for attrib in wireidentification:
@@ -341,7 +341,7 @@ class Harness:
             rows.append(["Additional components"])
             for extra in component.additional_components:
                 qty = extra.qty * component.get_qty_multiplier(extra.qty_multiplier)
-                if(self.mini_bom_mode):
+                if self.mini_bom_mode:
                     id = self.get_bom_index(extra.description, extra.unit, extra.manufacturer, extra.mpn, extra.pn)
                     rows.append(component_table_entry(f'#{id} ({extra.type.capitalize()})', qty, extra.unit))
                 else:
@@ -352,17 +352,15 @@ class Harness:
         bom_entries = []
         for part in component.additional_components:
             qty = part.qty * component.get_qty_multiplier(part.qty_multiplier)
-            bom_entries.append(
-                {
-                    'item': part.description,
-                    'qty': qty,
-                    'unit': part.unit,
-                    'manufacturer': part.manufacturer,
-                    'mpn': part.mpn,
-                    'pn': part.pn,
-                    'designators': component.name if component.show_name else None
-                }
-            )
+            bom_entries.append({
+                'item': part.description,
+                'qty': qty,
+                'unit': part.unit,
+                'manufacturer': part.manufacturer,
+                'mpn': part.mpn,
+                'pn': part.pn,
+                'designators': component.name if component.show_name else None
+            })
         return(bom_entries)
 
     def bom(self):
@@ -428,7 +426,6 @@ class Harness:
         bom_types_group = lambda bt: (bt['item'], bt['unit'], bt['manufacturer'], bt['mpn'], bt['pn'])
         for group in Counter([bom_types_group(v) for v in bom_entries]):
             group_entries = [v for v in bom_entries if bom_types_group(v) == group]
-            shared = group_entries[0]
             designators = []
             for group_entry in group_entries:
                 if group_entry.get('designators'):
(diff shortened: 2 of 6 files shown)


Tyler-Ward:
This is now just about functional but needs cleanup of the code, making the output look better, error checking, altered names to match #114 and addition of later sugestions in #50. Will continue this after #114 is done.
image

formatc1702:
Not too shabby!

My feeling is that the mini-BOMs within the nodes should be either one line per item (might stretch the nodes, I know),
one multiline cell per item (to easily distinguish separate items visually) or something similar. If I have a new suggestion, I can let you know.

Note: The unit is missing on the cable's sleeving, as well as the crimps and housing (should be showing `ea`)

formatc1702:
Another idea which might keep the diagram simple, but definitely require the BOM table, is the following:

- Auto-assign an Item Number (e.g. 1, 2, 3, ... or 10, 20, 30, ...) to every BOM item, show item number as first column in master BOM
- In the mini-BOMs, show one line = one cell per item in the format:
  `{qty} {unit} x {itemno}` for maximum minimalism
  or
  `{qty} {unit} x {itemno} ({description})` to keep it user-friendly but at least keeping the P/Ns out of it.

One could do a separate cell for each `{}` but hide the vertical borders within a line to get a nice table look without to much visual clutter?

formatc1702:
Quick reminder to use [imperative mood](https://chris.beams.io/posts/git-commit/#imperative) in git commit messages ;)

Tyler-Ward:
This branch has now been brought up to date with dev and the rest of the ideas from #50 have been aded in.

Merging the Bom items to a single cell definatley looks better. Using bom indexing to make the graph simpler sounds like a good idea, this will need some refactoring of the BOM generation code to get this to work but should be fairly easy to add once that is done. Other than that I think the only other thing that needs adding is adding a tutorial and syntax description (once #111 is merged in). 

image

Tyler-Ward:
The Bom has been refactored to remove duplications and add an id that can be looekd up when generating a graph. this is then used for the aditional components sections. The aditional component mode can be changed by a harness attribute to slect the option above but this will need #158 or similar  in order to be used

image

src/wireviz/Harness.py
```diff
@@ -394,24 +456,77 @@ def bom(self):
...
+                            'designators': cable.name
+                        }
+                    )
 
         for item in self.additional_bom_items:
             name = item['description'] if item.get('description', None) else ''
-            if isinstance(item.get('designators', None), List):
-                item['designators'].sort()  # sort designators if a list is provided
             item = {'item': name, 'qty': item.get('qty', None), 'unit': item.get('unit', None), 'designators': item.get('designators', None),
                     'manufacturer': item.get('manufacturer', None), 'mpn': item.get('mpn', None), 'pn': item.get('pn', None)}
```
kvid, line 491: I suggest removing all 7 `None` arguments in the 4 lines above. The second parameter of `.get()` is optional with default value`=None`. Simpler code is normally easier to read.

kvid (changes requested):
@Tyler-Ward Thank you for this very useful contribution. I'm just another contributor, like you. @formatc1702 is the owner that approves and merges PRs into dev when he thinks they are ready. He has been on vacation for several weeks, and I decided to review your PR and suggest some changes that I believe will help you getting this PR accepted, but remember that @formatc1702 might not share all my opinions.
- I like the BOM deduplication at the end as it enables removing a lot of other complex deduplication code.
- Consider moving more duplicated code into common (generic if needed) functions.

Tyler-Ward:
Thanks for your detailed comments @kvid. They all look good so have integrated them. Moving the components to a dataclass rather than a dict definatly helped the readability of the code in harness.py

src/wireviz/DataClasses.py
```diff
@@ -7,6 +7,22 @@
...
+@dataclass
+class AdditionalComponent:
+    type: str
+    subtype: Optional[str] = None
+    manufacturer: Optional[str] = None
+    mpn: Optional[str] = None
+    pn: Optional[str] = None
+    qty: float = 1
+    unit: Optional[str] = None
+    qty_multiplier: Optional[str] = None
```
kvid, line 19: Consider appending a comment to this line:
```py
# pincount | populated (for connectors), or wirecount | terminations | length | total_length (for cables)
```

kvid (changes requested):
- Thank you for accepting my suggestions.
- See code comments for some extra simplification suggestion

Tyler-Ward:
Thanks for the additional suggestions, Have added them as well, I left the swapping from dict to dataclass code the same as it was for now as the one liner is a bit harder to read but happy to swap to that if other people think that is better.

src/wireviz/Harness.py
```diff
@@ -97,8 +99,17 @@ def create_graph(self) -> Graph:
...
                      f'{connector.pincount}-pin' if connector.show_pincount else None,
                      connector.color, '<!-- colorbar -->' if connector.color else None],
-                    '<!-- connector table -->' if connector.style != 'simple' else None,
-                    [html_line_breaks(connector.notes)]]
+                    '<!-- connector table -->' if connector.style != 'simple' else None]
+            if connector.additional_components:
+                rows.append(["Additional components"])
+                for extra in connector.additional_components:
+                    qty = extra.qty * connector.get_qty_multiplier(extra.qty_multiplier)
+                    if(self.mini_bom_mode):
```
kvid, line 107: The usage of parentheses here is very unusual in Python. Maybe you're also a C/C++ programmer as I am? 😄 I suggest removing these parentheses from the `if`-statement.

src/wireviz/Harness.py
```diff
@@ -97,8 +99,17 @@ def create_graph(self) -> Graph:
...
+                        id = self.get_bom_index(extra.description, extra.unit, extra.manufacturer, extra.mpn, extra.pn)
+                        rows.append(component_table_entry(f'{id} ({extra.type.capitalize()})', qty, extra.unit))
```
kvid, line 109: Personally, I would prefer some prefix to the BOM id number in the component table for increased readability, 
e.g. `'#'`, `'No '` or `'BOM Id '`, but other users might disagree (e.g. also US/UK differences), and I suggest supporting an optional string value in the `mini_bom_mode` parameter in addition to the current True and False values. I therefore suggest inserting this code just below the current assignment of the `id` variable:
```py
if isinstance(self.mini_bom_mode, str):
    id = f'{self.mini_bom_mode}{id}'
```
The `if self.mini_bom_mode:` above will still work as intended for both `bool` and `str` values.

src/wireviz/Harness.py
```diff
@@ -330,90 +351,118 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
...
+            items = [v for v in bom_items if bom_types_group(v) == group]
             shared = items[0]
-            designators = [i['designator'] for i in items]
+            designators = []
+            for item in items:
+                if item.get('designators'):
+                    if isinstance(item['designators'], List):
+                        designators.extend(item['designators'])
+                    else:
+                        designators.append(item['designators'])
```
kvid, line 439: I suggest renaming `items` to `group_entries` and `item` to `inp` to avoid any confusion with the BOM `item` column name.

src/wireviz/Harness.py
```diff
@@ -330,90 +351,118 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
...
-            name = item['description'] if item.get('description', None) else ''
-            if isinstance(item.get('designators', None), List):
-                item['designators'].sort()  # sort designators if a list is provided
-            item = {'item': name, 'qty': item.get('qty', None), 'unit': item.get('unit', None), 'designators': item.get('designators', None),
-                    'manufacturer': item.get('manufacturer', None), 'mpn': item.get('mpn', None), 'pn': item.get('pn', None)}
-            bom_extra.append(item)
-        bom_extra = sorted(bom_extra, key=lambda k: k['item'])
-        bom.extend(bom_extra)
-        return bom
+            self._bom.append(item)
```
kvid, line 445: As above,  I suggest using `entry` as BOM entry variable.

src/wireviz/Harness.py
```diff
@@ -330,90 +351,118 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
...
+
+        self._bom = sorted(self._bom, key=lambda k: k['item'])  # sort list of dicts by their values (https://stackoverflow.com/a/73050)
+        # add index
+        index = 1
+        for item in self._bom:
+            item["id"] = index
```
kvid, line 451: As above,  I suggest using `entry` as BOM entry variable.

src/wireviz/Harness.py
```diff
@@ -330,90 +351,118 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
...
+            index += 1
+        return self._bom
+
+    def get_bom_index(self, item, unit, manufacturer, mpn, pn):
+        for bom_item in self.bom():
+            if((bom_item['item'], bom_item['unit'], bom_item['manufacturer'], bom_item['mpn'], bom_item['pn']) == (item, unit, manufacturer, mpn, pn)):
+                return bom_item['id']
```
kvid, line 458: As above,  I suggest using `entry` as BOM entry variable.

kvid (changes requested):
Thank you again for accepting most of my suggestions. I also like that you argue against suggestions that might harm readability. I'm sorry for suggesting more changes so many times, but I often find it hard to see all improvement issues at once. Some issues become much easier too see after some other improvement steps have been done first. I've marked some comments with DRY for [Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) because I prefer to avoid repeating the same code twice when it's possible to rewrite it to a loop, function call, or some other way to avoid code duplication without harming readability.

src/wireviz/Harness.py
```diff
@@ -330,90 +351,118 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
...
+                qty = part.qty * cable.get_qty_multiplier(part.qty_multiplier)
+                bom_items.append(
+                    {
+                        'item': part.description,
+                        'qty': qty,
+                        'unit': part.unit,
+                        'manufacturer': part.manufacturer,
+                        'mpn': part.mpn,
+                        'pn': part.pn,
+                        'designators': cable.name
```
kvid, line 419: I guess this code line should have ended with `if cable.show_name else None` as for connectors, but by using a common method, this should be fixed.

Tyler-Ward (commented):
No Problem, thanks for taking the time to keep reviewing. Some more good sugestions here, have implemented most of them and added comments on the others

kvid (changes requested):
Thank you once again for accepting most of my numerous suggestions. Please let me know when you are tired of improving details in this PR, and prefer raising a new issue to handle further improvements after merge-in, but please take a look at least at the two suggestions beginning with lines 441 and 448 in Harness.py.

_Edit:_ I'm really glad you've managed to both add this useful feature with additional components and that you in the process also simplified the BOM generating code that will be a great benefit to other PRs. The owner wrote in https://github.com/formatc1702/WireViz/issues/101#issuecomment-703774009 that he will finish the remaining PRs of v0.2 first, to avoid "feature creep", and then process the other PRs for v0.3. I suggested to him that your PR is a good candidate to start with, and hope he agrees when he arrives that stage. That means you should be prepared to rebase your PR branch on top of the new `dev` right after the v0.2 release. Thank you for all the good work!

formatc1702:
Progress on this PR looks fantastic, thank you guys for your heavy work so far.
@Tyler-Ward, I hope you're able to include @kvid's latest suggestions soon! 

Fair warning: I might have 1-2 little nit-picks as well before I merge, but @kvid has done the heavy lifting and overall, this looks very good!

src/wireviz/Harness.py
```diff
@@ -330,90 +351,118 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
...
+                    }
+                )
+
+        for item in self.additional_bom_items:
+            item = {'item': item.get('description', ''), 'qty': item.get('qty'), 'unit': item.get('unit'), 'designators': item.get('designators'),
+                    'manufacturer': item.get('manufacturer'), 'mpn': item.get('mpn'), 'pn': item.get('pn')}
+            bom_items.append(item)
+
```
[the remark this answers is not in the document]
Tyler-Ward: I had missed that the remove_line_breaks function would pass cleanly over non string items. so that would have worked fine.

I have done some tests on fields with newlines. Adding newlines (or any other whitespace changes) in fields will cause bom lines to be split into multiple rows. The catch with multiline fields is that by default the | operator in yaml appends a newline on the end generating additional spaces in the descriptions when run through remove_line_breaks() and causing them not to match due to that. 

There are a few options for how to deal with this.
1. Leave it as is with items being spit /defer this to a new issue.
2. Remove whitespace during deduplication matching, this could cause the unwanted whitespace to appear in the bom descriptions as we would need to pick one to use.
3. move the linebreak code back to the individual fields

I think 3 is probably the best soloution, will implement that tomorow unless there is a preference for one of the options.

Tyler-Ward (commented):
I have added the majority of the suggestions above, excluding the one that was sigested to skip and the one commented on below which has highlighted some bugs that need fixing which I will sort tomorrow.

> Thank you once again for accepting most of my numerous suggestions. Please let me know when you are tired of improving details in this PR, and prefer raising a new issue to handle further improvements after merge-in, but please take a look at least at the two suggestions beginning with lines 441 and 448 of Harness.py.

Thanks for the additional suggestions, Happy to keep on making edits now if there are further things to improve. or for them to be split into seperate tickets.

> Edit2: One of my code comments in this review is still tagged as Pending which I normally only see on comments before submitting the review, and I wonder if this comment was i fact submitted together with the rest of the review. Can you check if you see my comment as part of this review on line 343 of Harness.py with this contents:

That appears for me and has been fixed (along with a couple of others)

src/wireviz/Harness.py
```diff
@@ -329,91 +334,126 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
...
-        return bom
+        self._bom = sorted(self._bom, key=lambda k: k['item'])  # sort list of dicts by their values (https://stackoverflow.com/a/73050)
+
+        # add an incrementing id to each bom item
+        self._bom = [{**entry, 'id': index} for index, entry in enumerate(self._bom, 1)]
+        return self._bom
+
+    def get_bom_index(self, item, unit, manufacturer, mpn, pn):
+        for entry in self.bom():
+            if(entry['item'], entry['unit'], entry['manufacturer'], entry['mpn'], entry['pn']) == (item, unit, manufacturer, mpn, pn):
```
kvid, line 448:
```suggestion
            if (entry['item'], entry['unit'], entry['manufacturer'], entry['mpn'], entry['pn']) == tuple(clean_whitespace(v) for v in (item, unit, manufacturer, mpn, pn)):
```

src/wireviz/Harness.py
```diff
@@ -426,5 +466,6 @@ def bom_list(self):
             item_list = [item.get(key, '') for key in keys]  # fill missing values with blanks
             item_list = [', '.join(subitem) if isinstance(subitem, List) else subitem for subitem in item_list]  # convert any lists into comma separated strings
             item_list = ['' if subitem is None else subitem for subitem in item_list]  # if a field is missing for some (but not all) BOM items
+            item_list = [remove_line_breaks(subitem) for subitem in item_list]  # remove line breaks if present
```
kvid, line 469: If my other suggestion solves your problem, then this can be removed:
```suggestion
```

kvid (changes requested):
Thank you for accepting most of my suggestions again. Se my suggested solution to the problem you describe. I haven't had time to look at the other modifications you made yet.

Tyler-Ward:
That would catch some of the issues, the one I found that is easiest to end up in however is the following where a space is added before a comma. we could add code to replace any space comma combinations with a single comma, not sure if that would cause any other issues but can investigate that later today.

Below is a section of a bom where the connector type was split into two lines

Id | Item | Qty | Unit | Designators | P/N | Manufacturer | MPN
-- | -- | -- | -- | -- | -- | -- | --
3 | Connector, Molex Micro-Fit , female, 5 pins | 2 |   | X2, X3 | MF3.0-5 | Molex | 436450500
4 | Connector, Molex Micro-Fit, female, 5 pins | 1 |   | ZZ | MF3.0-5 | Molex | 436450500

kvid:
> That would catch some of the issues, the one I found that is easiest to end up in however is the following where a space is added before a comma. we could add code to replace any space comma combinations with a single comma, not sure if that would cause any other issues but can investigate that later today.

Please share  the YAML input that provoke the undesired output.

> 
> Below is a section of a bom where the connector type was split into two lines
> 
> Id	Item	Qty	Unit	Designators	P/N	Manufacturer	MPN
> 3	Connector, Molex Micro-Fit , female, 5 pins	2	 	X2, X3	MF3.0-5	Molex	436450500
> 4	Connector, Molex Micro-Fit, female, 5 pins	1	 	ZZ	MF3.0-5	Molex	436450500

I assume the first of these two entries was produced by specifying `additional_components` for X2 and X3. Am I right? In that case, the problem is caused by the comma added in `AdditionalComponent.description()`. I suggest calling `clean_whitespace()` on the attribute that is concatenated to the comma separator without a space between, i.e.:
```py
    @property
    def description(self) -> str:
        return clean_whitespace(self.type.capitalize()) + (f', {self.subtype}' if self.subtype else '')
```
I could not find any other code that concatenated a string attribute with a non-space separator before inserting into `Harness.bom_entries`.

Tyler-Ward:
The yaml that generated that bom is included at the end of the post (probably not the best test case  in its current form but can be turned into one).

The error is with the bom line for the connector itself (X2 and X3) where the type was split into two lines. ZZ is a aditional bom item that matches the output of a connector without a linebreak in its type. The extra space is added before the comma because the trailing newline is kept when the description string is built from the component variables (line 374 in the current state).

```yaml
connectors:
  X1:
    type: Powerpole
    pinlabels: [GND, VCC]
    ignore_in_bom: true
    additional_components:
      -
        type: housing
        subtype: Anderson Powerpole 15/45, Black
        qty: 1
        manufacturer: Anderson
        mpn: 1327G6
        pn: powerpole-black
      -
        type: housing
        subtype: Anderson Powerpole 15/45, Red
        qty: 1
        manufacturer: Anderson
        mpn: 1327
        pn: powerpole-red
      -
        type: crimp
        subtype: Anderson Powerpole 15/45, 30A
        qty_multiplier: pincount
        manufacturer: Anderson
        mpn: 1331
        pn: powerpole-crimp-30A
    notes: "notes go here"
  X2: &mf3-5
    type: |
      Molex
      Micro-Fit
    subtype: female
    pinlabels: [GND, VCC, A, B, C]
    mpn: 436450500
    manufacturer: Molex
    pn: MF3.0-5
    additional_components:
      -
        type: crimp
        qty_multiplier: populated
        manufacturer: Molex
        mpn: 43030-0007
        pn: MF3.0-Crimp

  X3:
    <<: *mf3-5

  F1:
    style: simple
    autogenerate: true
    type: Crimp ferrule
    subtype: 0.5 mm²
    color: OG # optional color
    additional_components:
      -
        type: crimp housing

cables:
    show_equiv: true
    length: 0.2
    mpn: 123456

  W2:
    <<: *wire_power
    additional_components:
      -
        type: |
            Sleve
        subtype: |
            Braided nylon,
            black, 3mm
        qty_multiplier: length
        unit: m
        pn: SLV-1

  W3:
    <<: *wire_power

connections:
  -
    - X1: [1-2]
    - W1: [1-2]
    - X2: [1-2]
  -
    - X1: [1-2]
    - W2: [1-2]
    - X3: [1-2]
  -
    - X1: [1-2]
    - W3: [1-2]
    - F1

additional_bom_items:
  - # define an additional item to add to the bill of materials (does not appear in graph)
    description: |
        Connector,
        Molex Micro-Fit, female, 5 pins
    qty: 1
    designators:
      - ZZ
    manufacturer: Molex
    mpn: 436450500
    pn: MF3.0-5
```

kvid:
> The yaml that generated that bom is included at the end of the post (probably not the best test case in its current form but can be turned into one).

Thank's for sharing. I believe it's a good candidate for future regression tests - you can suggest an improved version when you have the time.

> 
> The error is with the bom line for the connector itself (X2 and X3) where the type was split into two lines. ZZ is a aditional bom item that matches the output of a connector without a linebreak in its type. The extra space is added before the comma because the trailing newline is kept when the description string is built from the component variables (line 374 in the current state).

You are right. That means we need the same here: Calling `clean_whitespace()` separately on all the attribute that might be concatenated to the comma separator without a space between, i.e. `connector.type`, `connector.subtype`, and `cable.type`.

A better alternative is then to expand `clean_whitespace()` instead of these separate calls (including the separate call in the property method suggested earlier today):
```py
def clean_whitespace(inp):
    return ' '.join(inp.split()).replace(' ,', ',') if isinstance(inp, str) else inp
```
In any case, my other suggestions should still be valid (except the separate call in `AdditionalComponent.description()` if you accept my expanded `clean_whitespace()`).

I assume comma is the only non-space separator that can cause such problems...

Tyler-Ward:
I have looked through the fields that support newlines and in the bom they are all folowed by a comma or a space so that whitespace cleanup function will work and has been added.

I have rebased ontop of the latest dev to fix the merge conflict and also added the new syntax additions to the documentation.

src/wireviz/wv_helper.py
```diff
@@ -152,6 +152,9 @@ def graphviz_line_breaks(inp):
 def remove_line_breaks(inp):
     return inp.replace('\n', ' ').strip() if isinstance(inp, str) else inp
```
kvid, line 153: I suggest removing the no longer used functions `graphviz_line_breaks()` and `remove_line_breaks()` above.

kvid (changes requested):
Thank you for accepting some of my suggestions again, and for rebasing on top of the new `dev`.

tutorial/tutorial08.yml
```diff
@@ -28,6 +36,14 @@ cables:
     manufacturer: [WiresCo,WiresCo,WiresCo,WiresCo] # set a manufacter per wire
```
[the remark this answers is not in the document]
formatc1702: This is an issue that might have to be considered more generally, perhaps in the context of #56 and #127, where you might want other properties assigned on a per-wire basis.
Just something to keep in mind. I still think @kvid's suggestion could come in handy here.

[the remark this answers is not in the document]
kvid: My comment above is not about changing any feature. This is a feature that has been implemented several months ago and is present in v0.2. Wire attributes `pn`, `manufacturer`, and `mpn` can all be specified different for each wire in a list, or as a single value common to all wires. Please correct me if I'm wrong. See inserting [common values](https://github.com/formatc1702/WireViz/blob/v0.2/src/wireviz/Harness.py#L163-L166) and [list entries](https://github.com/formatc1702/WireViz/blob/v0.2/src/wireviz/Harness.py#L202-L206) in the diagram, and [inserting either](https://github.com/formatc1702/WireViz/blob/v0.2/src/wireviz/Harness.py#L379-L381) in the BOM.

My suggested comment in the tutorial is just to make the user aware of this feature, but please suggest a different/better way to demonstrate this feature. I'm sorry that the [doc](https://github.com/formatc1702/WireViz/blob/v0.2/docs/syntax.md#cable-attributes) slipped through my review without mentioning this feature.

If adding such a comment is a big issue that needs a discussion, then it's out of scope here. Adding a similar feature for other attributes is definitely out of scope here. 😃

src/wireviz/Harness.py
```diff
@@ -7,10 +7,11 @@
 from wireviz.wv_colors import get_color_hex
 from wireviz.wv_helper import awg_equiv, mm2_equiv, tuplelist2tsv, \
     nested_html_table, flatten2d, index_if_list, html_line_breaks, \
-    graphviz_line_breaks, remove_line_breaks, open_file_read, open_file_write, \
-    html_colorbar, html_image, html_caption, manufacturer_info_field
+    graphviz_line_breaks, remove_line_breaks, clean_whitespace, open_file_read, \
```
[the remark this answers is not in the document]
Tyler-Ward: I cant find any use of either of those functions in any other locations so is it woerth removing the functions as well?

src/wireviz/DataClasses.py
```diff
@@ -43,6 +46,21 @@ def __post_init__(self, gv_dir):
...
+    manufacturer: Optional[str] = None
+    mpn: Optional[str] = None
+    pn: Optional[str] = None
+    qty: float = 1
+    unit: Optional[str] = None
+    qty_multiplier: Union[ConnectorMultiplier, CableMultiplier, None] = None
+
+    @property
+    def description(self) -> str:
+        return self.type.capitalize().strip() + (f', {self.subtype.strip()}' if self.subtype else '')
```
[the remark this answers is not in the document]
Tyler-Ward: This is needed to avoid some graphicial odities in the diagram. Without the strip if either string is a multiline string (defined with | as in the examples) a traling newline is added which causes issues when the fields are joined together in the cell. Most of the other cells ignore this problem as the trailing <br/> is ignored by graphvis.

with mini bom
image
without mini bom
image

The alternatives could be either getting user to use the |- symbol instead (should update the docs in this case). or moving the cleanup of the "<br />,", "<br />)" and "<br /><br />" into the render code for the additional components cells.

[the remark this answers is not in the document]
formatc1702: In this case I'd argue having that little `.rstrip()` is nicer than requiring `|-` in the input YAML.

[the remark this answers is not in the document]
formatc1702: My question is, where does the need for `.capitalize()` come from?
Perhaps I'm missing something, but I don't see a need to mess with the user-defined `type` attribute. This applies to all instances of this function call.

src/wireviz/DataClasses.py
```diff
@@ -7,6 +7,9 @@
 from wireviz.wv_helper import int2tuple, aspect_ratio
 from wireviz import wv_colors
 
+ConnectorMultiplier = str  # = Literal['pincount', 'populated']
+CableMultiplier = str  # = Literal['wirecount', 'terminations', 'length', 'total_length']
+
```
formatc1702, line 12: Any reason the `Literal`s are commented out?

tutorial/tutorial08.yml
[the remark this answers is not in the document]
Tyler-Ward: Looking back at the code, there is actualy different functionality in the graph depening on if it is defined as a list or as a sting. If it is a string it will be listed in the top info of a bundle (like it is for a cable), or if it is a list it is listed per wire. The output can be altered so bundles are always rendered per wire even if it is a string, this is probably worth doing as it will avoid bundles looking incosistent.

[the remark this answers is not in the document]
formatc1702: @Tyler-Ward has been so kind and updated `syntax.md` to reflect the changes of this PR already, so I would request to add a note regarding this functionality (parameters accepting either `str` or `List[str]`) as well :)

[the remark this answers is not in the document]
formatc1702: Sorry, it seems we were typing a comment at the same time.

> The output can be altered so bundles are always rendered per wire even if it is a string, this is probably worth doing as it will avoid bundles looking incosistent.

Let's move that discussion to a new issue/PR so as not to delay this one too much, since it's drifting out of scope a bit. Based on my recent thoughts posted in #56 we'll need to think about the general issue of how to deal with parameters that differ between wires of a bundle, or perhaps even within a regular cable (e.g. different gauges).
So please don't worry about this point right now.

src/wireviz/DataClasses.py
Tyler-Ward: from @kvid's comment https://github.com/formatc1702/WireViz/pull/115#discussion_r497943873
> Note that the Literal type specifier is in comments to avoid requiring Python 3.8.

I have checked my 2.7 install and literals aernt supported, doing some searching I cant find any reference to this method of specifying literals so not sure if this is a standard implementation method or not.

formatc1702: Understood. Let's keep it like this then, perhaps add a comment behind the comment explaining this.

tutorial/tutorial08.yml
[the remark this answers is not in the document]
Tyler-Ward: Sounds sensible, will open an issue for this now.

src/wireviz/DataClasses.py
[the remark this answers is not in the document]
Tyler-Ward: The capitialise was originaly added to create more consistent bom formatting. We can remove this and pass that requirment onto the user if they care about it as this would be fairly intuative for a user to know what to change (unlike the newlines).

Tyler-Ward:
Thanks for the additional suggestions, I believe all of them have now been implemented or decided against in the inline comments but let me know if I have missed any.

src/wireviz/DataClasses.py
kvid: New in version 3.8 is specified by https://docs.python.org/3/library/typing.html#typing.Literal

src/wireviz/DataClasses.py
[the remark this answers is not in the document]
kvid: I agree with your decision to remove the `.capitalize()`. Personally, I think I would have preferred recommending `|-` in the doc before removing the trailing newline with `.rstrip()`, but I understand that others might have a different opinion. The only motivation for a user to use `|` when specifying this attribute is to insert newlines, and then it shouldn't be too much of a surprise seeing the results above. If we in the doc write "recommend `|-` to avoid a trailing newline", that should be enough, IMHO, but I respect that you disagree. No hard feelings. 😄

formatc1702:
Once there are no further comments, I will likely squash this PR into a single commit, unless @Tyler-Ward wants to go through the trouble (not worth it IMHO) of compressing the current 30 commits into something more manageable with meaningful intermediate steps.
@kvid since you've been keeping an eye on the implementation, I will rely on your GO to proceed with the squash+merge.

kvid (approved):
Thank you for accepting more suggestions. As far as I can see now, you have either implemented or decided against all my suggestions. I provided a link to the documentation of `Literal` that you couldn't find in that code comment thread, and explained why I would prefer recommending `|-` before calling `.rstrip()`, but that I also respect your opinion.

Thank you for the fantastic effort to develop this PR with the useful feature and improving the BOM code in the process. I'm also impressed by your patience to process my numerous suggestions about details.

PR #116, opened 2020-07-21
[refactor] Use the csv library to generate properly escaped TSV files

### LINKED ISSUE #98: [feature] CSV BOM output
While the TSV type output is fine, if one is directly ordering from the BOM, such as through Digikey, CSV format is preferred. A simple way to do this is to switch the custom tsv generation to the CSV library, which supports both tsv and CSV, and using additional flags to determine which formats are outputted.

Additionally (and this may be best as a separate issue), the SVG output may be able to be in-lined with the HTML so that it can be an option that can be turned on and off depending on what the user wants, just like with the different BOM formats.

aakatz3:
This closes #98 and allows escaping tabs in TSV files. It also uses library functions to generate the CSV and TSV files, instead of a quick custom solution, for robustness.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -3,9 +3,9 @@
 
 from wireviz.DataClasses import Connector, Cable
 from graphviz import Graph
-from wireviz import wv_colors, wv_helper
+from wireviz import wv_colors, wv_helper, bom_helper
 from wireviz.wv_colors import get_color_hex
-from wireviz.wv_helper import awg_equiv, mm2_equiv, tuplelist2tsv, \
+from wireviz.wv_helper import awg_equiv, mm2_equiv, \
     nested_html_table, flatten2d, index_if_list, html_line_breaks, \
     graphviz_line_breaks, remove_line_breaks, open_file_read, open_file_write
 from collections import Counter
@@ -297,8 +297,8 @@ class Harness:
         graph.save(filename=f'{filename}.gv')
         # bom output
         bom_list = self.bom_list()
-        with open_file_write(f'{filename}.bom.tsv') as file:
-            file.write(tuplelist2tsv(bom_list))
+        # todo: support user choices of BOM format (probably also graphviz outputs, html outputs)
+        bom_helper.generate_bom_outputs(filename,bom_list,bom_helper.WIREVIZ_TSV, bom_helper.EXCEL_CSV)
         # HTML output
         with open_file_write(f'{filename}.html') as file:
             file.write('<!DOCTYPE html>\n')
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -298,7 +298,7 @@ class Harness:
         # bom output
         bom_list = self.bom_list()
         # todo: support user choices of BOM format (probably also graphviz outputs, html outputs)
-        bom_helper.generate_bom_outputs(filename,bom_list,bom_helper.WIREVIZ_TSV, bom_helper.EXCEL_CSV)
+        bom_helper.generate_bom_outputs(filename,bom_list, [bom_helper.WIREVIZ_TSV, bom_helper.EXCEL_CSV])
         # HTML output
         with open_file_write(f'{filename}.html') as file:
             file.write('<!DOCTYPE html>\n')
--- a/src/wireviz/bom_helper.py
+++ b/src/wireviz/bom_helper.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import csv
+from wireviz import wv_helper
+from wireviz.wv_helper import open_file_write
+
+EXCEL_CSV = csv.excel
+EXCEL_TSV = csv.excel_tab
+UNIX_CSV = csv.unix_dialect
+WIREVIZ_TSV = type('Wireviz BOM', (csv.Dialect, object), dict(
+    delimiter='\t',
+    doublequote=True,
+    escapechar=None,
+    lineterminator='\n',
+    quoting=0,
+    skipinitialspace=False,
+    strict=False,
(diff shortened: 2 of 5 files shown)


formatc1702:
Thanks, will check it out later. Looks like a good addition.
Did you confirm the TSV output remains the same as it previously was?

aakatz3:
TSV output matches exactly, other than fixing a bug where if a tab was present inside of any of the data fields, it was previously not escaped. Now, the field is quoted. If you would like, I can restore the original behavior.

formatc1702:
Great that it matches exactly! And even better if it escapes existing tabs, that's much better. Let me test it a bit and I'll merge.

src/wireviz/bom_helper.py
```diff
@@ -0,0 +1,42 @@
...
+    expanded_tsv_names = len(_tsv_formats.intersection(set(argv))) > 1
+    for fmt in argv:
+        if fmt in _csv_formats:
+            file = csv.writer(open_file_write(base_filename + ("_" + fmt.__name__ if expanded_csv_names else "") + _csv_ext, fmt.lineterminator), fmt)
+
+        elif fmt in _tsv_formats:
+            file = csv.writer(open_file_write(base_filename + ("_"+fmt.__name__ if expanded_tsv_names else "") + _tsv_ext, fmt.lineterminator), fmt)
+        else:
+            raise KeyError("Unknown BOM Format Specified")
+        file.writerows(wv_helper.flatten2d(bomdata))
```
formatc1702, line 40: The `*argv` parsing (lines 29 to 31) are not very pythonic, I'd request you change it.
The `len(intersection(set())) > 1` is rather cryptic... and the use case of calling this function requesting multiple CSV and TSV formats at once is pretty much an edge case IMHO... especially since currently there are set defaults, inaccessible to the user.

My proposal: Call the function with a list as the third argument:
```
def generate_bom_outputs(base_filename, bomdata, formats):
    if not isinstance(formats, List):
        formats = [formats]
```
and then iterate over the list.
It could then be called using
```
bom_helper.generate_bom_outputs(filename,bom_list,[bom_helper.WIREVIZ_TSV, bom_helper.EXCEL_CSV])
```

src/wireviz/bom_helper.py
```diff
@@ -0,0 +1,42 @@
...
+
+# TODO: Possibly refactor other BOM output operations, such as HTML, into here?
```
formatc1702, line 42: Yes, but in a separate PR :)

src/wireviz/Harness.py
```diff
@@ -297,8 +297,8 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
         graph.save(filename=f'{filename}.gv')
         # bom output
         bom_list = self.bom_list()
-        with open_file_write(f'{filename}.bom.tsv') as file:
-            file.write(tuplelist2tsv(bom_list))
+        # todo: support user choices of BOM format (probably also graphviz outputs, html outputs)
+        bom_helper.generate_bom_outputs(filename,bom_list,bom_helper.WIREVIZ_TSV, bom_helper.EXCEL_CSV)
```
formatc1702, line 301: See comment in `bom_helper.py`.

src/wireviz/wv_helper.py
```diff
@@ -113,8 +113,8 @@ def remove_line_breaks(inp):
     return inp.replace('\n', ' ').rstrip() if isinstance(inp, str) else inp
 
 def open_file_read(filename):
-    # TODO: Intelligently determine encoding
+    # TODO: Intelligently determine encoding (UnicodeDammit, Chardet, cchardet are not very reliable in testing)
```
formatc1702, line 116: This change is unnecessary XD

src/wireviz/bom_helper.py
formatc1702: See the discussion in #123

src/wireviz/bom_helper.py
aakatz3: That could work well! I will also check briefly to see if the Excel TSV format is essentially compatible with the one wireviz natively generates (and probably have the OS choose if new lines are unix or windows style), and then the csv formats can be directly passed through.

aakatz3: Also, excuse any python style issues, I will try to fix them whenever pointed out. I'm a Java / C / MATLAB developer, primarily, who has wanted to learn python for far too long!

formatc1702: > I'm a Java / C / MATLAB developer, primarily, who has wanted to learn python for far too long!

You're not alone, I've done mostly embedded programming in C++ / Arduino before.
I'm loving Python's readability and trying to keep WireViz as pythonic and understandable as possible :)

src/wireviz/bom_helper.py
aakatz3: Understood, leaving comment for future reference, unless requested otherwise.

src/wireviz/Harness.py
aakatz3: Understood, leaving comment for future reference, unless requested otherwise.

formatc1702:
Please see my [proposal](https://github.com/formatc1702/WireViz/pull/123#issuecomment-663598361) in #123.

I encourage to submit this pull request to @slightlynybbled's fork, as he's willing to integrate the new CSV/TSV output into a refactored version that implements a clean CLI and better separation of wiring harness logic from file output.

Please comment there and close this PR if you are in.

aakatz3:
Will do, resolving your issues first then opening new pull request.

formatc1702:
Please make sure to submit the new PR to [slightlynybbled's fork](https://github.com/slightlynybbled/wireviz/pulls), not here. That way he can take care of integrating it, and then submit it together with his CLI improvements onto the original WireViz repo.

PR #121, opened 2020-07-22
Feature/short bom names

Tyler-Ward:
This implements #114 commit b97e0ea implements the medged manufacturer and mpn fields examples of each are shown below this can be included or excluded from the merge depending on peoples opinion on which version looks better.

Seperated fields:
ex02

Merged fields:
ex02


formatc1702:
Looks good! You have my vote for the second option (merged fields).
I'll try it out as soon as I can.

One minor thing: since we've moved P/N in front of the manufacturer info, let's do the same in the BOM, for consistency.
`PN | Manuf. | MPN`

Also, the exported BOM header still has the long names, please shorten.

Thanks!

src/wireviz/wv_helper.py
```diff
@@ -117,4 +117,12 @@ def open_file_read(filename):
...
\ No newline at end of file
+    return open(filename, 'w', encoding='UTF-8')
+
+
+def manufacturer_info_field(manufacturer, mpn):
+    if manufacturer:
+        part_number_component = f': {mpn}' if mpn else ''
+        return(f'{manufacturer}{part_number_component}')
+    else:
+        return(f'MPN: {mpn}' if mpn else None)
```
formatc1702, line 128: How about this?
```python
if manufacturer or mpn:
  return f'{manufacturer if manufacturer else "MPN"}{": " if mpn else ""}{mpn if mpn else ""}'
else
  return None
```

It should return the same four options:
`None` (nothing specified)
`{manufacturer}` (no MPN specified)
`MPN: {mpn}` (no manufacturer specified)
`{manufacturer}: {mpn}` (both specified)

kvid: How about this?
```
    if manufacturer:
        return f'{manufacturer}: {mpn}' if mpn else manufacturer
    else:
        return f'MPN: {mpn}' if mpn else None
```

Tyler-Ward: That looks like a much tider way to implement that, have swapped to that implementation in the latest commits

Tyler-Ward:
The changes in the above comments have been added, let me know if there are any other edits wanted.

PR #128, opened 2020-07-26
default should be an iterable

gstein:
The code looks at `fmt` as an iterable. Tweak the default value to avoid iteration of 'p', 'd', 'f'.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -211,7 +211,7 @@ class Harness:
 
         return dot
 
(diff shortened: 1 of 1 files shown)


formatc1702:
Sorry, this PR is based on the month-old `master` release.
It has been fixed in `dev` and there is ongoing refactoring work, so this is no longer required.
Thanks for your input, nonetheless!

gstein:
Ah! Whoops. I'll rebase off dev, then.

Issue #129, opened 2020-07-27
[internal] Improve readability of manufacturer_info_field()


kvid:
During the review of #121, I suggested:
@formatc1702 How about this alternative instead?
```py
    if manufacturer:
        return f'{manufacturer}: {mpn}' if mpn else manufacturer
    else:
        return f'MPN: {mpn}' if mpn else None
```
Edit: It's closer to the original, except without a variable and without [parentheses around the return expressions](https://stackoverflow.com/questions/4978567/should-a-return-statement-have-parentheses).

_Originally posted by @kvid in https://github.com/formatc1702/WireViz/pull/121#discussion_r459378362_

@Tyler-Ward seemed to like my suggestion:
> That looks like a much tider way to implement that, have swapped to that implementation in the latest commits

_Originally posted by @Tyler-Ward in https://github.com/formatc1702/WireViz/pull/121#discussion_r460534914_

However, it was not part of the merge into `dev`, and I therefore rise the suggestion again.

formatc1702:
Sorry, it was a bit unclear which variant @Tyler-Ward's response was referring to, and I took

> have swapped to that implementation in the latest commits

to mean that the PR already contained his preferred choice.

Personally, I stand behind my proposal, since it separates the "show the MPN cell or not?" logic in the first stage (`if` statement) from the "how to format the cell?" logic in the second stage (`return` statement), and is thus cleaner IMO.
I would argue this is down to personal coding style and not a huge deal-breaker either way :)
 
@Tyler-Ward, please let us know if you'd like to change it, since it's your code contribution and I'm happy to oblige.

Tyler-Ward:
> > have swapped to that implementation in the latest commits
> 
> to mean that the PR already contained his preferred choice. Perhaps I misinterpreted this.

That is correct, while the suggestion from @kvid looks cleaner when reading the code. I felt that the sugestion from @formatc1702 made the the fact that it returned either a string or None clearer.

kvid:
> That looks like a much tider way to implement that, have swapped to that implementation in the latest commits

> > > have swapped to that implementation in the latest commits
> > 
> > 
> > to mean that the PR already contained his preferred choice. Perhaps I misinterpreted this.
> 
> That is correct, while the suggestion from @kvid looks cleaner when reading the code. I felt that the sugestion from @formatc1702 made the the fact that it returned either a string or None clearer.

[...]

PR #130, opened 2020-07-27
Improve readability of manufacturer_info_field()

kvid:
Fixes issue #129, but originally suggested in review of #121.


kvid:
Issue #129 was based on an unfortunate misunderstanding.

PR #118, opened 2020-07-21
Extended build examples

kvid:
This PR ~~builds on top of #110 and~~ add some new features:
- Add `compare` action to compare generated files (except those generated by Graphviz unless using `-c`) against git repository.
- Add `restore` action to restore generated files from git repository.
- Make all actions honor the optional argument `-g`. This make it possible to select the group of files to use for any action by following the action argument with `-g` and then the space separated group names.

Edit: It should not break the existing `build` and `clean` actions, but their algorithms to collect files are slightly adjusted so they can use a common function to reduce code duplication.

These new features are useful for #63, but does not close it (collecting corner cases are also needed).

Edit 2: A few extra changes seen by the user:
- The CLI option `-generate` is renamed to `-g`/`--groups` and some argument help is added.
- The group name is included in the status output as the path is equal for two groups, and now it contains `{description} {groupkey} in {path}` e.g. `Building demos in /path/WireViz/examples`.

Usage:
- Run `python build_examples.py` to build generated files in all groups.
- Run `python build_examples.py compare` to compare generated files in all groups against the last commit.
- Run `python build_examples.py clean` to delete generated files in all groups.
- Run `python build_examples.py restore` to restore generated files in all groups from the last commit.
- Edit 4: Append `-c` or `--compare-graphviz-output` to the `compare` command above to also compare the Graphviz output (default: False).
- Append `-g` or `--groups` followed by space separated group names to any command above, and the set of generated files affected by the command will be limited to the selected groups.
- Run `python build_examples.py -h` or with `--help` to print the generated help. The auto-generated help text is not the best, but better than maintaining it manually. I wonder if [click](https://click.palletsprojects.com/en/7.x/) as recommended in https://github.com/formatc1702/WireViz/issues/60#issuecomment-663278159 could improve this?

Edit 3: This PR was originally built on top of #110, but is now rebased onto `dev` to prepare merge-in. The single commit from #110 that my commits depend on, is therefore now included here. Some of my commits are squashed together. The original branch before squash-rebasing is available as [kvid:extended-build-examples-before-rebase](https://github.com/kvid/WireViz/tree/extended-build-examples-before-rebase).

Personally, I want to use `python build_examples.py` followed by `python build_examples.py compare` as a regression test to easily see what has changed in the output files just re-generated compared to the latest commit. In addition, I want to use  `python build_examples.py restore` just before committing to avoid unwanted changes from e.g. a different Graphviz version. Hopefully, others also find this useful.

Issue #43, opened 2020-06-29
How to install correctly after refactoring?


kvid:
Yesterday morning, I was running WireViz (latest release at that time) and created a small pull request with a proposed feature enhancement. Since then, a major refactoring was done, and I now need to use the latest release to update my pull request. That was not so easy as I hoped.

1. First, I tried using my existing repo, checking out master branch and pulling the newest release from this public repo. That went OK.

1. I realized the embedded Graphviz classes was gone, and tried `pip3 install -e .` from the main folder as described in the [README](https://github.com/formatc1702/WireViz/blob/master/README.md), but still I got `ModuleNotFoundError: No module named 'graphviz'`.

1. I tried `sudo apt-get install python-apt` as suggested at [stackoverflow](https://stackoverflow.com/questions/13708180/python-dev-installation-error-importerror-no-module-named-apt-pkg) since I also got the `ModuleNotFoundError: No module named 'apt_pkg'` error, but still no luck.

1. I tried `sudo apt-get install graphviz libgraphviz-dev pkg-config` as I found in a [check log for a recent pull request](https://github.com/formatc1702/WireViz/pull/39/checks), but still no luck.

1. I tried `pip3 install graphviz` as described [here](https://pypi.org/project/graphviz/), but still no luck.

1. I tried creating a new clean local repo, and install again, but still no luck:
```
git clone https://github.com/formatc1702/WireViz.git
cd WireViz/
pip3 install -e .
cd src/wireviz
python3.7 build_examples.py
```
Here is the full Traceback:
```
Traceback (most recent call last):
  File "build_examples.py", line 8, in <module>
    from wireviz import wireviz
  File "../wireviz/wireviz.py", line 6, in <module>
    from graphviz import Graph
ModuleNotFoundError: No module named 'graphviz'
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 63, in apport_excepthook
    from apport.fileutils import likely_packaged, get_recent_crashes
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 30, in <module>
    import apport.fileutils
  File "/usr/lib/python3/dist-packages/apport/fileutils.py", line 23, in <module>
    from apport.packaging_impl import impl as packaging
[...]

PR #133, opened 2020-07-27
Feature/update readme

formatc1702:
I've restructured the [main README file](https://github.com/formatc1702/WireViz/blob/feature/update-readme/README.md), added and moved around some of the information. I welcome any new input to make the readme more user-friendly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 # WireViz
 
+
 ## Summary
 
 WireViz is a tool for easily documenting cables, wiring harnesses and connector pinouts. It takes plain text, YAML-formatted files as input and produces beautiful graphical output (SVG, PNG, ...) thanks to [GraphViz](https://www.graphviz.org/). It handles automatic BOM (Bill of Materials) creation and has a lot of extra features.
+
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


formatc1702:
A better description for the `build_examples.py` options is probably a good thing to add.

README.md
```diff
@@ -90,8 +81,39 @@ Output file:
...
+
+#### Installing the development version
+
+Access to the current state of the development branch can be gained by cloning the repo and installing manually:
+
+```
+git clone <repo url>
+cd <working copy>
+git checkout dev
+pip3 install -e .
```
formatc1702, line 112: Please make sure this workflow makes sense.

kvid:
> A better description for the `build_examples.py` options is probably a good thing to add.

See PR #118 description for a short summary in the Usage section, and a couple of use cases in the last section, to get some ideas.

formatc1702:
It might make sense to create a `docs/` directory or similar, for the changelog, the syntax description, and the `build_examples.py` documentation. This way we don't overload the readme, or the parent directory.
In the future, this can be replaced by a properly built documentation system.

README.md
kvid: Should you also mention something like the alternative variant you gave to me in https://github.com/formatc1702/WireViz/issues/43#issuecomment-651078861?

```sh
git clone https://github.com/formatc1702/WireViz.git
cd WireViz/
python -m venv wv-env
source wv-env/bin/activate
pip3 install -e .
cd src/wireviz
python build_examples.py
```

Note that `git checkout dev` is missing above, and I have not tried any of these installations from scratch recently.

kvid: Should there be a reference to CONTRIBUTING.md in the section about Installing the development version?

formatc1702:
I'm merging this into `feature/syntax-description` since they are closely related; let's continue the discussion there.

README.md
formatc1702: Sorry, didn't see these comments earlier.
- I'm not sure whether we should give advice on virtual environments directly, or reference a site that explains the concept more generally.
- Sure, I can include a link to `CONTRIBUTING.md`

PR #140, opened 2020-07-30
Bugfix/dash in pinlabel

### LINKED ISSUE #139: [bug] Referencing pin labels with '-' when defining connections
Discovered on a modified version of the example from #138.

When using the `-` character in a pin label, and then referencing this pin label in `connections`, the character is interpreted to be a range of pins, rather than part of the pin label itself.

```yaml
connectors:
  Molex1:
    type: automotive
    subtype: male
    pincount: 26
    pins: [19,20,21,22,23]
    pinlabels: [GA-NET+, AMP, RCA+, GA-NET-, RCA-]
    notes: GA-NET(BUS) has 68 ohm 1/2 watt resistor

cables:
  Cable1:
    wirecount: 5
    color_code: DIN

connections:
  -
    - Molex1: [GA-NET+, GA-NET-]
    - Cable1: [1-2]
```

In the second to last line, `GA-NET+` is interpreted as a range from `GA` to `NET+`, and an error is thrown.
```
ValueError: invalid literal for int() with base 10: 'GA'
```

The example works when replacing all `-` in `pinlabels` with `_` or other.

The parsing should be modified to check for existing labels first, before calling `wv_helper.expand()`
OR
`wv_helper.expand()` should be made smarter.

formatc1702:
Fixes #139.

### DIFF
--- a/src/wireviz/wv_helper.py
+++ b/src/wireviz/wv_helper.py
@@ -59,21 +59,26 @@ def expand(yaml_data):
         yaml_data = [yaml_data]
     for e in yaml_data:
         e = str(e)
-        if '-' in e:  # list of pins
-            a, b = tuple(map(int, e.split('-')))
-            if a < b:
-                for x in range(a, b + 1):
-                    output.append(x)
-            elif a > b:
-                for x in range(a, b - 1, -1):
-                    output.append(x)
-            elif a == b:
-                output.append(a)
+        if '-' in e:
+            a, b = e.split('-', 1)
+            try:
+                a = int(a)
+                b = int(b)
+                if a < b:
+                    for x in range(a, b + 1):
+                        output.append(x)  # ascending range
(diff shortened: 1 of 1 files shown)


src/wireviz/wv_helper.py
```diff
@@ -59,21 +59,26 @@ def expand(yaml_data):
...
+            try:
+                a = int(a)
+                b = int(b)
+                if a < b:
+                    for x in range(a, b + 1):
+                        output.append(x)  # ascending range
+                elif a > b:
+                    for x in range(a, b - 1, -1):
+                        output.append(x)  # descending range
+                elif a == b:
```
kvid, line 73: This line can be replaced by a simple `else:` AFAIK.

formatc1702: Agreed.

PR #150, opened 2020-08-12
fix #149 introduced in #147

### LINKED ISSUE #149: [bug] Rendering multi-colored striped wires has changed in dev due to merging an unfinished PR
Bugs:
- Two-colored wires are only rendered with two colors, not with the first color repeated after the second color, as before.
- Single-colored wires in the same harness as multi-colored wires are not expanded to equalize the wire thickness, as before.

A fix is described in my change request https://github.com/formatc1702/WireViz/pull/147#discussion_r465883313 (acknowledged with 👍 by @formatc1702) of the last PR review before merging, but implementing the fix was probably forgotten before the merge.

See also [dev branch commit history](https://github.com/formatc1702/WireViz/commits/dev).

SnowMB:
I did not come around implementing the fix by @kvid in time and now we have a regression regarding padding colors. This should be the needed fix.

### DIFF
--- a/src/wireviz/wv_colors.py
+++ b/src/wireviz/wv_colors.py
@@ -120,7 +120,7 @@ def get_color_hex(input, pad=False):
         padded = input
 
     try:
-        output = [_color_hex[padded[i:i + 2]] for i in range(0, len(input), 2)]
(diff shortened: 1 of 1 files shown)


formatc1702:
Sorry, I misread the code review comment and assumed it was already fixed. My bad.

PR #137, opened 2020-07-30
Add optional image and caption attributes to Connector and Cable/Bundle nodes

kvid:
- This fixes issue #27.
- Edit: All scaling attributes described in https://github.com/formatc1702/WireViz/issues/27#issuecomment-670141605 are now implemented.

Issue #151, opened 2020-08-13
[internal] Create a new module for helper functions to each output format?


kvid:
As the number of helper functions for generating HTML is increasing, maybe we should move them into a new `wv_html.py` file? See also the excerpt from a recent discussion below.

> > > Do you have a set of rules or recommendations about where to insert newlines, i.e. what HTML tags can be on the same row, and when should they be separated?
> > 
> > 
> > No hard rules right now, and it's not terribly important. Cells are usually in one row (`<td>...</td>`). Multiple short opening tags in sequence tend to go together (`<tr><td>` and the respective closing tags for the outermost table, for example). Row breaks (`</tr><tr>`) could go in one line but are usually split because they come from iterations of a loop.
> 
> Nice. Should we but these words into a comment in the code as soft recommendations?
> 
> Maybe into `wv_helper.py`, or into a new `wv_html.py` and move the increasing number (at least in my PR #137) of html related helper functions there too?

A new module for the HTML stuff sounds good, as a separate issue/PR!

> > > Have you considered a single space indentation for each sub-level of HTML tag structure?
> > 
> > 
> > ~I thought about it but decided against it; the gains are marginal, the readability of the Python code would suffer; plus it's easy to pass the HTML through some auto-formatter afterwards if one really needs to analyze the structure at some point.~
> > Edit: thinking about it some more, it might not be such a bad idea, I'll try it out and push a commit.
> 
> It is not a major issue to me, especially if the Python code gets more complicated and less readable. I will wait and see what you suggest.

It was less painful than anticipated, and I do like the output! Have a look at the latest commit. This will hopefully be the last change before the merge.

_Originally posted by @formatc1702 in https://github.com/formatc1702/WireViz/pull/136#issuecomment-673109385_

formatc1702:
I recently got a similar feeling regarding @Tyler-Ward's BOM generation. (See additional thoughts below)
Before I open a new issue, I suggest widening the scope of this one, to something like "split functions into separate modules" or similar.
Either way, I suggest opening a PR that only moves functions around and only modifies code where absolutely necessary (e.g. passing an object instead of referring to `self`).
If @kvid or @Tyler-Ward want to take care of this, let me know; otherwise, I can do it myself as well.

---

Perhaps it would be best to move the BOM functions (currently in `Harness.py`, as functions inside the `Harness` class) to a separate `wv_bom.py`. The functions would then get passed a `Harness` object instead of `self`.
Candidates include:
[...]

PR #153, opened 2020-08-13
Add optional image to connectors and cables

### LINKED ISSUE #27: Include connector pinout diagrams
This would be a fantastic feature, but might be hard to implement.

Some questions
* Load existing graphics or auto-generate (how?)?
* One diagram per connector series/family/type, or one for every specific part?
* From which side of connector? Cable side or mating side?

kvid:
Add optional image to connectors and cables (#153)

This image, with an optional caption below, is displayed in the lower 
section of the connector/cable node in the diagram - just above the 
notes if present.

This solves the basic part of issue #27, and is a continuation of 
PR #137 that was closed due to changes in the base branch.


kvid:
Scaling can be tested e.g. like this for a smaller image:
```diff
------------------------------ examples/ex08.yml ------------------------------
index cd67a29..01c6329 100644
@@ -18,6 +18,8 @@ cables:
     wirecount: 3
     shield: SN
     image: ../resources/cable-WH+BN+GN+shield.png
+    image_scale: true
+    image_size: [50, 50, fixed]
     caption: Cross-section
 
 connections:
```

src/wireviz/DataClasses.py
```diff
@@ -18,6 +18,10 @@ class Connector:
     type: Optional[str] = None
     subtype: Optional[str] = None
     pincount: Optional[int] = None
+    image: Optional[str] = None
+    image_scale: Optional[str] = "FALSE"
```
formatc1702, line 22: why not make this a boolean?

src/wireviz/Harness.py
```diff
@@ -96,15 +96,13 @@ def create_graph(self) -> Graph:
                     [html_line_breaks(connector.type),
                      html_line_breaks(connector.subtype),
                      f'{connector.pincount}-pin' if connector.show_pincount else None,
-                     connector.color, '<!-- colorbar -->' if connector.color else None],
+                     connector.color, html_colorbar(connector.color)],
```
formatc1702, line 99: I am not against this change, but it probably should be a separate PR since it is unrelated to the image issue

src/wireviz/DataClasses.py
kvid: Because it has several valid values as described in https://github.com/formatc1702/WireViz/issues/27#issuecomment-670141605.

src/wireviz/Harness.py
```diff
@@ -171,15 +169,13 @@ def create_graph(self) -> Graph:
                      f'{cable.gauge} {cable.gauge_unit}{awg_fmt}' if cable.gauge else None,
                      '+ S' if cable.shield else None,
                      f'{cable.length} m' if cable.length > 0 else None,
-                     cable.color, '<!-- colorbar -->' if cable.color else None],
+                     cable.color, html_colorbar(cable.color)],
                     '<!-- wire table -->',
+                    [html_image(cable)],
+                    [html_caption(cable)],
```
formatc1702, line 175: `html_caption(cable.caption)` is nicer than passing along the entire cable object for a single property

src/wireviz/Harness.py
kvid: I agree. I put it here to demonstrate the power of the new functionality.

src/wireviz/Harness.py
kvid: The function need more than one attribute from the node.

src/wireviz/Harness.py
```diff
@@ -171,15 +169,13 @@ def create_graph(self) -> Graph:
                      f'{cable.gauge} {cable.gauge_unit}{awg_fmt}' if cable.gauge else None,
                      '+ S' if cable.shield else None,
                      f'{cable.length} m' if cable.length > 0 else None,
-                     cable.color, '<!-- colorbar -->' if cable.color else None],
+                     cable.color, html_colorbar(cable.color)],
                     '<!-- wire table -->',
+                    [html_image(cable)],
```
formatc1702, line 174: `html_image(cable.image, cable.image_scale, cable.image_size)` maybe?

src/wireviz/wv_helper.py
```diff
@@ -53,6 +56,41 @@ def nested_html_table(rows):
...
 
+def html_colorbar(color):
+    return f'<tdX bgcolor="{wv_colors.translate_color(color, "HEX")}" width="4">' if color else None
+
+def html_image(node):
+    if not node.image:
+        return None
+    size = html_size_attr(node.image_size)
+    # The leading attributes belong to the preceeding tag. See where used below.
+    html = f'{"".join(size)}><img scale="{node.image_scale}" src="{node.image}"/>'
```
formatc1702, line 67: why do `"".join(size)` here? why not make `html_size_attr()` directly return a string that can be included as-is?

src/wireviz/wv_helper.py
```diff
@@ -53,6 +56,41 @@ def nested_html_table(rows):
...
+    # - First value is minimum width of object in points, an int value in the range 1-65535. (Default 0 = none)
+    # - Second value is minimum height of the object in points, an int value in the range 1-65535. (Default 0 = none)
+    # - Third value is a boolean where TRUE specify that the first two values are the fixed image cell size.
+    #   (Default FALSE) Other string values, e.g. FIXED are also interpreted as TRUE.
+    if not size or not isinstance(size, list):
+        return [''] * 3
+    return [f' width="{size[0]}"'  if len(size) > 0 and size[0] else '',
+            f' height="{size[1]}"' if len(size) > 1 and size[1] else '',
+             ' fixedsize="TRUE"'   if len(size) > 2 and size[2] else '']
+
```
formatc1702, line 93: See comment above.
Just return an empty string (line 89) or do a `"".join()` before returning.

examples/ex08.yml
```diff
@@ -7,14 +7,19 @@ connectors:
     pins: [T, R, S]
     pinlabels: [Dot, Dash, Ground]
     show_pincount: false
+    image: ../resources/stereo-phone-plug-TRS.png
```
formatc1702, line 10: Please move the `resources/` directory into the `examples/` directory, it shouldn't be in the repo's root.

src/wireviz/Harness.py
kvid: It also needs the `cable.caption`, and I see no harm to let the helper function see all common attributes of nodes. In object oriented code, this would have been a method of the node superclass.

src/wireviz/DataClasses.py
```diff
@@ -18,6 +18,10 @@ class Connector:
...
+    image_size: List[Any] = field(default_factory=list)
```
formatc1702, line 23: Does it make sense to have a list here, instead of separate `image_width` and `image_height` attributes of type `int`? That removes the guessing of what order the list is (width or height first).

src/wireviz/wv_helper.py
```diff
@@ -43,7 +45,8 @@ def nested_html_table(rows):
                 html.append('  <table border="0" cellspacing="0" cellpadding="3" cellborder="1"><tr>')
                 for cell in row:
                     if cell is not None:
-                        html.append(f'   <td balign="left">{cell}</td>')
+                        # Inject attributes to the preceeding <td> tag where needed
+                        html.append(f'   <td balign="left">{cell}</td>'.replace('><tdX', ''))
```
formatc1702, line 49: the `<tdX` hack is clever, but I'm wondering if there is a more transparent way to show what's going on? I don't have a specific suggestion at this point, but will think about it.

src/wireviz/wv_helper.py
kvid: Because I also need to test if it has a `fixedsize` attribute set, and I then I would need to test if the resulting string ends with `fixedsize="TRUE"` by repeating this string is two different functions or defining it as a global string - both resulting in less readable code, IMHO. A third option is to repeat the test of the input list, but that is repeating the same code twice.

src/wireviz/wv_helper.py
```diff
@@ -53,6 +56,41 @@ def nested_html_table(rows):
...
+def html_caption(node):
+    return f'''<tdX{' sides="LRB"' if node.image else ''}>{html_line_breaks(node.caption)}''' if node.caption else None
+
+def html_size_attr(size):
+    # Return Graphviz HTML attributes to specify minimum size of a TABLE or TD object
+    # size: List of values where only these are used:
+    # - First value is minimum width of object in points, an int value in the range 1-65535. (Default 0 = none)
+    # - Second value is minimum height of the object in points, an int value in the range 1-65535. (Default 0 = none)
+    # - Third value is a boolean where TRUE specify that the first two values are the fixed image cell size.
+    #   (Default FALSE) Other string values, e.g. FIXED are also interpreted as TRUE.
```
formatc1702, line 87: Does such a small function require such a long comment? The goal should be to have code that is understandable enough that most comments are unnecessary, maybe there is potential for simplification here... The int range is interesting, but maybe irrelevant since it's unlikely somebody will really need an image 65535 px wide, so it's a tradeoff between documenting for edge cases, vs. readability.
What do you think?

src/wireviz/DataClasses.py
formatc1702: I see. In the YAML, maybe we can use lowercase strings (`false`, `true`, `width`) to keep it consistent, and the helper function can take care to convert `false` and `true` (which the YAML parser will interpret as boolean) to strings, and make them uppercase before including in the HTML.

examples/ex08.yml
kvid: I explained the reason and asked your opinion in https://github.com/formatc1702/WireViz/issues/27#issuecomment-668704773:
> I suggest storing images for embedding in a new resources folder adjacent to examples. Then it is easier to access the same folder also from a tutorial if that might be needed in the future. Do you rather prefer a folder below examples?

Do you think it is OK to access `../examples/resources/image.png` from a tutorial or is it never useful to access the same images from both examples and tutorial YAML files? I can do the change, as long as you understand the implications I'm thinking about.

src/wireviz/Harness.py
formatc1702: How about nesting the image attributes in the input YAML file?
```yaml
connectors:
  Key:
    type: Phone Connector
    subtype: male 3.5
    pins: [T, R, S]
    pinlabels: [Dot, Dash, Ground]
    show_pincount: false
    image:
      src: ../resources/stereo-phone-plug-TRS.png
      caption: Tip, Ring, and Sleeve
      width: 100
      height: 100
      scale: both
```

- It keeps all the attributes neatly together
- You only need to pass the `node.image` attribute to the function and have all you need
- results in shorter attributes (`scale`, `width`, ... instead of `image_scale`, `image_size`, ...)

src/wireviz/wv_helper.py
kvid: The value type, limit, and default and are copied from Graphviz doc. Should we rather link to that?

formatc1702: A small note to the GraphViz doc should be enough, yes, it's a good compromise :)

formatc1702: PS: Please don't mark conversations as "resolved" when there are open questions, otherwise they might go unnoticed. I only resolve when there is no disagreement at all and the change is implemented exactly according to the comment.

examples/ex08.yml
formatc1702: If necessary, I see no harm in having both `examples/resources/` and `tutorial/resources` directories if necessary, to avoid having to do any `../` directory traversing. So my answer is yes, I do prefer a folder inside `examples` (and another one inside `tutorials` if needed)

src/wireviz/wv_helper.py
kvid: It's just a generalization of the `.replace()` from your code version. The selected tag name `<tdX>` is a bit cryptic and could be spelled out to e.g. `<join-with-previous-tag>`, but that would result in longer code lines that might harm readability. With all functions using this collected in a new module as suggested in #151, that would be a place to explain this implementation technique.

src/wireviz/wv_helper.py
kvid: A fourth option is to always include the extra nested table and then not needing the test if it has a `fixedsize` attribute set in this function.

formatc1702: Why do you need to check the output of `html_size_attr()` to know if it has the `fixedsize` attribute, instead of getting it from the image attributes [as I suggest](https://github.com/formatc1702/WireViz/pull/153/files/9a6522395913788c6583aeee96342203fe007cbe#r470489101)

src/wireviz/wv_helper.py
kvid: I did return a single string in the earlier commit https://github.com/formatc1702/WireViz/pull/153/commits/41b3f3acfedb42dacf600f81985ef67547b58e37, but changed it to avoid the extra nested table when I could avoid it in the next commit. See also my comment to the other code fragment above.

src/wireviz/wv_helper.py
formatc1702: > A fourth option is to always include the extra nested table and then not needing the test if it has a `fixedsize` attribute set in this function.

This is also an option, since the extra table doesn't really hurt anybody :-)

src/wireviz/DataClasses.py
kvid: I can do that if that's preferred. I just read somewhere that it's more Pythonic to specify a width x height size as a tuple, but I'm not sure how to specify a tuple in the YAML input.

formatc1702: AFAIK there are no tuples in YAML. And expecially having the`fixed` parameter in there (which might be a bit cryptic for a regular user) is not very pythonic IMHO, so having an easy to read syntax [as suggested](https://github.com/formatc1702/WireViz/pull/153#discussion_r470489101) is probably the user-friendliest compromise.

The whole width x height tuple sounds like something out of Numpy, or OpenCV maybe, but leaving out the guesswork for the user is preferrable IMO.

src/wireviz/DataClasses.py
kvid: Graphviz is not case sensitive. We can use upper, lower, or capital case. The Graphviz doc uses uppercase on all tags and attributes, but lowercase seems to be the standard in this project. The YAML parser interpretes both `true` and `TRUE` as boolean, and the f-string result is `True` in both cases, but Graphviz don't care. I can change to lowercase in comments and constant strings, but do we also need `str(value).lower()` in the f-string when Graphviz don't really care?

formatc1702: [As you wish 😎](https://www.youtube.com/watch?v=CxK_nA2iVXw)

src/wireviz/Harness.py
kvid: Good idea! I like it, and I will look into what changes are needed.

src/wireviz/wv_helper.py
formatc1702: Agreed. We can leave it as it is now, and clean up / improve / explain in #151 as you suggest.

src/wireviz/Harness.py
formatc1702: I am open to changing `src` to `source`, `file`, `filename` or anything else if you think it's more user friendly... but `src` is also fine by me.

examples/ex08.yml
kvid: I will do as requested. I have drawn a 4-pin Molex KK 254 connector house that could have been useful in both places, though.
MolexKK254_4-pin
I'm sorry for nagging. I forgot to include this image example in my previous comment. I did draw it as SVG, but my Graphviz installation doesn't accept embedded SVG images. 😞

formatc1702: Even still, I think it's OK to have two instances of the tiny image file (one in `examples/resources/` and one in `tutorial/resources/`), if that means fewer `../`s in the path :)

src/wireviz/wv_helper.py
kvid: A new `image` class might change this picture completely.

src/wireviz/wv_helper.py
kvid: With a new `image` class, maybe this can be reversed back to returning a simple string.

src/wireviz/wv_helper.py
formatc1702: Just to make sure: Are you planning on implementing a true `image` class? Or just have an `image` attribute of `dict` type, with `src`, `caption`, `width` etc. as keys? That should probably be enough, right?

kvid: I prefer a true Image dataclass to obtain the easier syntax `image.src` instead of `image['src']`. The only processing needed are two extra lines in `__post_init__()` of the two classes containing an image.

src/wireviz/DataClasses.py
```diff
@@ -7,6 +7,16 @@
 from wireviz import wv_colors
 
 
+@dataclass
+class Image:
+    src: str
+    scale: str = "false"
```
formatc1702, line 13: Please add a comment to show the possible values... no explanation (that's something for the `docs`), just a simple cheat sheet:
```python
    scale: str = "false"  #  false | true | width | height | both
```

formatc1702:
I'm very happy with how this came out! See my only comment regarding a little cheat sheet for the [not so obvious] `scale` attribute.

Since the end result of this PR is very straightforward (mainly just adding lines, not many changes to existing ones), I would like to squash it all into one commit if that is OK with you. Please let me know! Thanks a lot for this contribution.

src/wireviz/DataClasses.py
```diff
@@ -7,6 +7,16 @@
...
+    width: Optional[int] = None
+    height: Optional[int] = None
+    fixedsize: bool = False
```
formatc1702, line 16: Shouldn't this be `Optional` as well?

src/wireviz/DataClasses.py
formatc1702: And probably make the attribute `Optional`... it's not used in `ex08.yml`, for example

formatc1702:
Please see my comments in 89b730b16c93c655e53a23f23517bb994610bf59 regarding the documentation.

src/wireviz/DataClasses.py
kvid: > And probably make the attribute `Optional`... it's not used in `ex08.yml`, for example

It is optional, but I just followed the pattern from the other optional bool attributes:
```
    show_name: bool = None
    show_pincount: bool = None
    hide_disconnected_pins: bool = False
    autogenerate: bool = False
```

src/wireviz/DataClasses.py
kvid: See https://github.com/formatc1702/WireViz/pull/153#discussion_r471052220

src/wireviz/DataClasses.py
formatc1702: TBH I am not sure what effect the `Optional[]` keyword has.
Is there a difference between the following two lines?
```python
    scale: str = 'false'
    scale: Optional[str] = 'false'
```
I'm having trouble finding an answer, and currently both versions are used in `DataClasses.py` so I would like to decide on one for everything :)

kvid:
> Scaling can be tested e.g. like this for scaling the plug uniformly within a fixed 100x100 cell (obtaining a smaller image), and scaling the cable to fill available space in both directions within a cell with 200 as minimum height (obtaining a larger and distorted image):

The same scale test with new syntax:
```diff
----------------------------- examples/ex08.yml ------------------------------
index d4299f0..f8c2248 100644
@@ -9,6 +9,10 @@ connectors:
     show_pincount: false
     image:
       src: resources/stereo-phone-plug-TRS.png
+      scale: true
+      width: 100
+      height: 100
+      fixedsize: true
       caption: Tip, Ring, and Sleeve
 
 cables:
@@ -21,6 +25,8 @@ cables:
     shield: SN
     image:
       src: resources/cable-WH+BN+GN+shield.png
+      scale: both
+      height: 300
       caption: Cross-section
 
 connections:
```

Other use cases might be:
- Only create more space around an image using (`width` x `height`) > natural size of image, and no scaling.
- Enlarge image in available horizontal space using `scale: width`.
- Enlarge image uniformly in available space using `scale: true` (and `height` > natural height of image to make space for uniform scaling).

Do you think we need a separate example to demonstrate scaling?

formatc1702:
> Do you think we need a separate example to demonstrate scaling?

I think one example showing basic usage of images is enough at the moment.
I suggest we add this feature to `dev` now, and think how we can make things easier for the user in the next iteration.

Reason:
I can't shake the feeling that all these parameters will cause more confusion than benefit for 99% of users.
A WireViz user should not need to worry about the internal workings of the GraphViz table and image parameters, IMHO.
From a user point of view, the required functions would be:
- Insert an image and caption
- Set width OR height to scale the image proportionally, if necessary
- Set width AND height to stretch the image to a specific size and ratio, in rare exceptions

Therefore my proposed solution is this:
Get the PR merged now, and released in v0.2... this way, users can try it out and give valuable feedback, in the shape of bug reports, feature requests, etc.
For the next release (v0.3), we might think of ways to simplify the user experience as I mention above.
What do you think?

kvid:
> > Do you think we need a separate example to demonstrate scaling?
> 
> I think one example showing basic usage of images is enough at the moment.

I agree. Including an image is not a major functionality of WireViz.

> I suggest we add this feature to `dev` now as it is, and think how we can make things easier for the user in the next iteration.

OK, unless you also want my minor suggestions described below.

> 
> Reason:
> I can't shake the feeling that all these parameters will cause more confusion than benefit for 99% of users.
> A WireViz user should not need to worry about the internal workings of the GraphViz table and image parameters, IMHO.

If I'm optimistic, maybe 5 % of the WireViz users might want to include an image at least once. Of those users that want to include an image, I guess 90 % only need `src` and `caption`. The remaining 10 % might consider scaling because:
1. Their unscaled image appear too large and expand the node too much.
2. Their unscaled image appear much smaller than what can fit in the node.
3. Very seldom, their unscaled image appear distorted.

> From a user point of view, the required functions would be:
> 
> * Insert an image and caption

That's the trivial part. It's already supported. 😃

To keep this functionality, we can compute the default values of the unspecified attributes:
- `scale` = "false" if `width` and `height` are both unspecified.
- `fixedsize` = False if `width` and `height` are both unspecified.

> * Set width OR height to scale the image proportionally, if necessary

To obtain this, we can compute the default values of the unspecified attributes:
- Find the image aspect ratio and compute the missing `width` or `height`.
- `scale` = "true" if either `width` or `height` is specified, but not both.
- `fixedsize` = True if `width` or `height` is specified.

> * Set width AND height to stretch the image to a specific size and ratio, in rare exceptions

To obtain this, we can compute the default values of the unspecified attributes:
- `scale` = "both" if both `width` and `height` are specified.
- `fixedsize` = True if `width` or `height` is specified.

> 
> (this is how the regular HTML `` tag's `HEIGHT` and `WIDTH` attributes behave, for example)
> Users should not care/worry about the `scale` and `fixedsize` parameters, should they?

If we can compute sensible default values in most cases, then the users only need to specify them in very rare cases.

The only non-trivial issue above is to find the image aspect ratio. I'm not sure if it's possible to obtain this from a Graphviz preprocessing pass or if we need to open the image, find the type , and then for any type supported by Graphviz, find the aspect ratio in the file header. If possible, this is a task for v0.3, I guess.

I suggest, we assume a 1:1 aspect ratio, for now, and make `width` and `height` equal when any of them are not specified. The result will still be a proportionally scaled image, but within a fixed square image cell. When the smallest dimension is missing, the effect is extra image cell space in that dimension. When the largest dimension is missing, the effect is a smaller image than expected. If this is not good enough for the user, it's always possible to solve it by specifying both dimensions.

It's still possible for an advanced user to specify all attributes, but I assume the computed default values are good enough in almost all use cases.

> 
> Therefore my proposed solution is this:
> Get the PR merged now, and released in v0.2... this way, users can try it out and give valuable feedback, in the shape of bug reports, feature requests, etc.
> For the next release (v0.3), we might think of ways to simplify the user experience as I mention above.
> What do you think?

I agree, but consider also my minor suggestions described above. See also my latest commit.

> 
> Please don't think I am criticizing your implementation! I only want to make sure we don't lose focus of the simplicity of the WireViz syntax, thinking only from a developer's perspective :)

No, I don't feel criticized. Suggestions that make it easier for the users are welcome.

src/wireviz/DataClasses.py
```diff
@@ -11,15 +11,30 @@
...
     height: Optional[int] = None
-    fixedsize: Optional[bool] = False
+    fixedsize: Optional[bool] = None
     # Contents of the cell <td> just below the image cell:
     caption: Optional[str] = None
     # See also HTML doc at https://graphviz.org/doc/info/shapes.html#html
 
+    def __post_init__(self):
+
+        if self.scale is None:
```
formatc1702, line 25: `if self.scale:`?

src/wireviz/DataClasses.py
```diff
@@ -11,15 +11,30 @@
...
+            self.scale = "false" if self.width is None     and self.height is None \
+                else     "both"  if self.width is not None and self.height is not None \
+                else     "true" # When only one dimension is specified.
+
+        if self.fixedsize is None:
+            self.fixedsize = self.width is not None or self.height is not None
+
+        if self.width is None and self.height is not None:
```
formatc1702, line 33: maybe `if self.width and not self.height:` is enough, and a bit more readable?
Since `width` and `height` are both `int`, only a value of 0 should yield false, in which case this would probably also be the appropriate behavior.
Same for line 35 of course.

src/wireviz/DataClasses.py
```diff
@@ -11,15 +11,30 @@
...
     # Contents of the cell <td> just below the image cell:
     caption: Optional[str] = None
     # See also HTML doc at https://graphviz.org/doc/info/shapes.html#html
 
+    def __post_init__(self):
+
+        if self.scale is None:
+            self.scale = "false" if self.width is None     and self.height is None \
+                else     "both"  if self.width is not None and self.height is not None \
+                else     "true" # When only one dimension is specified.
```
formatc1702, line 28: while `self.scale` and `self.fixedsize` need the explicit `is None`, `width` and `height` do not IMHO, see comment below.

src/wireviz/wv_helper.py
```diff
@@ -75,8 +75,8 @@ def html_caption(image):
 
 def html_size_attr(image):
     # Return Graphviz HTML attributes to specify minimum or fixed size of a TABLE or TD object
-    return ((f' width="{image.width}"'   if image.width else '')
-        +   (f' height="{image.height}"' if image.height else '')
+    return ((f' width="{image.width}"'   if image.width is not None else '')
+        +   (f' height="{image.height}"' if image.height is not None else '')
```
formatc1702, line 79: See comment above, this change is probably not needed?

formatc1702:
> > I can't shake the feeling that all these parameters will cause more confusion than benefit for 99% of users.
> > A WireViz user should not need to worry about the internal workings of the GraphViz table and image parameters, IMHO.
> 
> If I'm optimistic, maybe 5 % of the WireViz users might want to include an image at least once. Of those users that want to include an image, I guess 90 % only need `src` and `caption`. The remaining 10 % might consider scaling [...]

Agree. I meant 99% of users needing an image won't care about the `scale` and `fixedsize` parameters directly.

Your implementation looks very reasonable.

> If we can compute sensible default values in most cases, then the users only need to specify them in very rare cases.
> 
> The only non-trivial issue above is to find the image aspect ratio. I'm not sure if it's possible to obtain this from a Graphviz preprocessing pass or if we need to open the image, find the type , and then for any type supported by Graphviz, find the aspect ratio in the file header. If possible, this is a task for v0.3, I guess.

Yes, let's leave that for later. My hope is that we won't need to load the image and find out the aspect ratio, but rather can use some combination of attributes (`scale="height"` maybe) that will do what we want.

> I suggest, we assume a 1:1 aspect ratio, for now, and make `width` and `height` equal when any of them are not specified.

👍 

I like your changes very much (see my minor code comments).
I won't be able to test this today but I would consider this PR complete once you accept or reject my latest comments.
Thank you for your effort!

src/wireviz/DataClasses.py
```diff
@@ -7,6 +7,35 @@
...
+
+    def __post_init__(self):
+
+        if self.scale is None:
+            self.scale = "false" if self.width is None     and self.height is None \
+                else     "both"  if self.width is not None and self.height is not None \
+                else     "true" # When only one dimension is specified.
+
+        if self.fixedsize is None:
+            self.fixedsize = self.width is not None or self.height is not None
```
formatc1702, line 31: See comment below

kvid:
> Your implementation looks very reasonable.

Thank's

> 
> > If we can compute sensible default values in most cases, then the users only need to specify them in very rare cases.
> > The only non-trivial issue above is to find the image aspect ratio. I'm not sure if it's possible to obtain this from a Graphviz preprocessing pass or if we need to open the image, find the type , and then for any type supported by Graphviz, find the aspect ratio in the file header. If possible, this is a task for v0.3, I guess.
> 
> Yes, let's leave that for later.

Alternatives (both with and without an external library) to retrieve image width and height are discussed here: https://stackoverflow.com/questions/8032642/how-to-obtain-image-size-using-standard-python-class-without-using-external-lib

I can understand you are not very keen on adding another library dependency, but the non-library alternative looks ugly and might be a challenge to maintain for all supported image types. What do you think? Should I call the PIL (Python Image Library) if it is installed, or assume 1:1 aspect ratio otherwise?
```py
def calc_missing_dimension(image_src, dim, width_is_missing):
    try:
        from PIL import Image
        im = Image.open(image_src)
        return dim*im.width/im.height if width_is_missing else dim*im.height/im.width
    except ModuleNotFoundError as err:
        return dim # Assuming 1:1 aspect ratio.
```
The code above is working as long as I use absolute path to the image. What is the easiest way to translate a relative image path inside a YAML file to the correct absolute path?

> My hope is that we won't need to load the image and find out the aspect ratio, but rather can use some combination of attributes (`scale="height"` maybe) that will do what we want.

I'm not aware of any attribute combination for scaling to a fixed size, other than `fixedsize=True`, and since that also requires both `width` and `height` to be specified, we need the aspect ratio of the image to avoid the already described down-sides when assuming 1:1 aspect ratio. AFAIK, any attribute combination with `fixedsize=False` can only expand the image cell size, and that does not comply with your specification.

Maybe we need to submit a PR to Graphviz that allows `fixedsize=True` with only one dimension specified when containing an image, and then Graphviz can calculate the missing dimension using the aspect ratio, but I guess it can take quite a while to get such a PR accepted, then wait for the next release, and then wait for e.g. the OS packet manager to include the new release.

> 
> > I suggest, we assume a 1:1 aspect ratio, for now, and make `width` and `height` equal when any of them are not specified.
> 
> 👍

OK

> 
> I like your changes very much (see my minor code comments).

About `if image.width is not None` vs `if image.width`, I wasn't sure if a zero value could be useful in any combination, and was therefore reluctant to silently block that by interpreting it as not specified.

Today, I had a look into the Graphviz source code and found the function `size_html_cell()` that check the `width` and `height` values. This code outputs a warning about unspecified `width` or `height` when any of them are zero:
  https://gitlab.com/graphviz/graphviz/-/blob/master/lib/common/htmltable.c#L1210-1221

This probably means you are right about a zero `width` or `height` value is regarded equal as if it was not specified, and therefore it should be safe to only test `if image.width` instead of `if image.width is not None`. I'm not 100 % sure, because I've only looked into a fraction of the Graphviz source code, but it's a strong indicator. Do you have more knowledge about this subject?

Just below in the same function, you can also see that the cell space can only expand when `fixedsize=False`:
  https://gitlab.com/graphviz/graphviz/-/blob/master/lib/common/htmltable.c#L1223-1224

> I won't be able to test this today but I would consider this PR complete once you accept or reject my latest comments.

I might have to wait until tomorrow by committing and pushing the last changes.

> Thank you for your effort!

Thank's for the review.

src/wireviz/wv_helper.py
kvid: You probably mean: _replaced by `if image.X`?_ (like it was earlier)

src/wireviz/DataClasses.py
kvid: You probably mean: _`X is not None` VS `X`_

kvid:
I just found a small bug. I will force-push a fix into the last two commits. Can you wait an hour with any review?

src/wireviz/DataClasses.py
```diff
@@ -28,14 +28,17 @@ def __post_init__(self):
...
-            self.fixedsize = self.width or self.height
-
-        if self.height:
-            if not self.width:
-                self.width = self.height # Assuming 1:1 aspect ratio for now.
-        else:
-            if self.width:
-                self.height = self.width # Assuming 1:1 aspect ratio for now.
+            # Normally equal to (self.width or self.height), but the user might override self.scale.
+            self.fixedsize = str(self.scale).lower() in ["true", "both"]
```
kvid, line 34: I also found a draw-back with the new default value: The `fixedsize` might now be set default True even if no dimension is specified. That will result in an error from Graphviz. I should change it to a combination of the two conditions to avoid that:
```
# Avoid default True if the user overrides self.scale to a different value.
self.fixedsize = (self.width or self.height) and str(self.scale).lower() in ["true", "both"]
```

src/wireviz/DataClasses.py
```diff
@@ -3,10 +3,46 @@
...
 from dataclasses import dataclass, field
-from wireviz.wv_helper import int2tuple
+from pathlib import Path
+from wireviz.wv_helper import int2tuple, aspect_ratio
 from wireviz import wv_colors
 
 
+@dataclass
+class Image:
+    gv_dir: Path # Directory of .gv file injected as context during parsing
```
kvid, line 13: Maybe the type should be `InitVar[Path]` because it is [only used during initialization](https://docs.python.org/3/library/dataclasses.html#init-only-variables) in the `__post_init__()` function?

martonmiklos:
Hi @kvid !

I did a test drive with this PR and ran to the following minor issue:
I have tried to use an image source which filename contained a space and it made the generation failing:

```
mm@mm-precision:~/Projektek/machine_cabling/MDB cable$ ../WireViz/src/wireviz/wireviz.py MDB\ cable.yaml 
Warning: "MDB cable.png" was not found as a file or as a shape library member
Error: No or improper image file="MDB cable.png"
in label of node W1
  File "../WireViz/src/wireviz/wireviz.py", line 257, in <module>
    main()
  File "../WireViz/src/wireviz/wireviz.py", line 253, in main
    parse(yaml_input, file_out=file_out)
  File "../WireViz/src/wireviz/wireviz.py", line 183, in parse
    harness.output(filename=file_out, fmt=('png', 'svg'), view=False)
  File "../WireViz/src/wireviz/../wireviz/Harness.py", line 298, in output
    graph.render(filename=filename, view=view, cleanup=cleanup)
  File "/home/mm/.local/lib/python3.8/site-packages/graphviz/files.py", line 207, in render
    rendered = backend.render(self._engine, format, filepath,
  File "/home/mm/.local/lib/python3.8/site-packages/graphviz/backend.py", line 221, in render
    run(cmd, capture_output=True, cwd=cwd, check=True, quiet=quiet)
  File "/home/mm/.local/lib/python3.8/site-packages/graphviz/backend.py", line 183, in run
    raise CalledProcessError(proc.returncode, cmd,
graphviz.backend.CalledProcessError: Command '['dot', '-Tsvg', '-O', 'MDB cable']' returned non-zero exit status 1. [stderr: b'Warning: "MDB cable.png" was not found as a file or as a shape library member\nError: No or improper image file="MDB cable.png"\nin label of node W1\n']
```

Here is the relevant part of the yaml:

```
cables:
  W1:
    gauge: 16 AWG
    length: 1
    color_code: DIN
    wirecount: 5
    shield: false
    colors: [RD, BK, YE, GN, BU, GR]
    image: 
        src: 'MDB_cable.png'
        width: 100
        height: 100
```

I think enclosing the image sources when calling the dot could solve this issue.

kvid:
> Hi @kvid !
> 
> I did a test drive with this PR and ran to the following minor issue:

Thank's for testing my PR.

> I have tried to use an image source which filename contained a space and it made the generation failing:
> 
> ```
> mm@mm-precision:~/Projektek/machine_cabling/MDB cable$ ../WireViz/src/wireviz/wireviz.py MDB\ cable.yaml 
> Warning: "MDB cable.png" was not found as a file or as a shape library member
> Error: No or improper image file="MDB cable.png"
> in label of node W1
> Traceback (most recent call last):
>   File "../WireViz/src/wireviz/wireviz.py", line 257, in <module>
>     main()
>   File "../WireViz/src/wireviz/wireviz.py", line 253, in main
>     parse(yaml_input, file_out=file_out)
>   File "../WireViz/src/wireviz/wireviz.py", line 183, in parse
>     harness.output(filename=file_out, fmt=('png', 'svg'), view=False)
>   File "../WireViz/src/wireviz/../wireviz/Harness.py", line 298, in output
>     graph.render(filename=filename, view=view, cleanup=cleanup)
>   File "/home/mm/.local/lib/python3.8/site-packages/graphviz/files.py", line 207, in render
>     rendered = backend.render(self._engine, format, filepath,
>   File "/home/mm/.local/lib/python3.8/site-packages/graphviz/backend.py", line 221, in render
>     run(cmd, capture_output=True, cwd=cwd, check=True, quiet=quiet)
>   File "/home/mm/.local/lib/python3.8/site-packages/graphviz/backend.py", line 183, in run
>     raise CalledProcessError(proc.returncode, cmd,
> graphviz.backend.CalledProcessError: Command '['dot', '-Tsvg', '-O', 'MDB cable']' returned non-zero exit status 1. [stderr: b'Warning: "MDB cable.png" was not found as a file or as a shape library member\nError: No or improper image file="MDB cable.png"\nin label of node W1\n']
> ```

Your error above tells me WireViz was called with `MDB\ cable.yaml` as input and when calling Graphviz with `MDB\ cable.gv` as input to generate `MDB\ cable.svg` and `MDB\ cable.png` it failed while trying to read "MDB cable.png" as an embedded image. It seems Graphviz is instructed to output the same image file that also is embedded. That will not work.

> 
> Here is the relevant part of the yaml:
> 
> ```
> cables:
>   W1:
>     gauge: 16 AWG
>     length: 1
>     color_code: DIN
>     wirecount: 5
>     shield: false
>     colors: [RD, BK, YE, GN, BU, GR]
>     image: 
>         src: 'MDB_cable.png'
>         width: 100
>         height: 100
> ```

In the YAML input above the embedded image is 'MDB_cable.png' with an underscore, but that does not fit with the error above where Graphviz is trying to embed "MDB cable.png" with a space.

I have tried your YAML above with a copy of the cable image from example 08 renamed to 'MDB_cable.png' and I'm unable to reconstruct your reported error with this YAML input. However, if I put `src: 'MDB cable.png'` with a space in the YAML file, then the `MDB\ cable.png` is overwritten and the error occur, but that's expected as you cannot embed the same image file as Graphviz is writing to at the same time.

> 
> I think enclosing the image sources when calling the dot could solve this issue.

If you mean enclosing the filename in quotes, that is already done. Take a look in the generated `.gv` file and you see what is input to Graphviz.

Please verify that you embed a different image file than the output image file, and also verify that the image file you embed still is a valid image and not an empty file because it might have been overwritten by Graphviz in one of your early experiments.

martonmiklos:
You are right, I just wanted to make a quick test and using the output file was a quick but bad idea.

kvid:
> Sorry, for personal reasons I was not able to work on the project this week, and now I won't have time to test your new feature until after my holidays (ending in two weeks). I look forward to trying this out.
> I don't want to rush things so I will try out your code once I have time! :)

Don't worry. You don't have to promise anything. You decide when things are ready for release. There are always things in our lives (like family, friends, etc.) that are far more important that a software project.

> 
> > I can understand you are not very keen on adding another library dependency, but the non-library alternative looks ugly and might be a challenge to maintain for all supported image types. What do you think? Should I call the PIL (Python Image Library) if it is installed, or assume 1:1 aspect ratio otherwise?
> 
> Using PIL is fine.

I had already pushed the `aspect_ratio()` function that uses PIL, but fall back to 1:1 aspect ratio if failing for any reason.

> My only request right now would be to add `PIL` to `requirements.txt` and `setup.py` as a new dependency.

Are you sure you want to require PIL to be installed? My current implementation catch the `ModuleNotFoundError` if PIL is not found and fall back to 1:1 aspect ratio in such a case.

> 
> > The code above is working as long as I use absolute path to the image. What is the easiest way to translate a relative image path inside a YAML file to the correct absolute path?
> 
> Maybe get the YAML's absolute path and interpret the image path as relative to that? Can't give a code example right now but [`pathlib`](https://realpython.com/python-pathlib/) should make this relatively painless.

I had already done that. I just needed a clean way to pass the directory of the `.gv` file into the `Image` dataclass, and solved that by injecting it as a value in the `yaml_data`.

> 
> > This probably means you are right about a zero `width` or `height` value is regarded equal as if it was not specified, and therefore it should be safe to only test `if image.width` instead of `if image.width is not None`. I'm not 100 % sure, because I've only looked into a fraction of the Graphviz source code, but it's a strong indicator. Do you have more knowledge about this subject?
> 
> No idea, I simply assumed that 0 didn't make sense ;-) It was just a guess.

I decided to trust the strong indicator I found and pushed your suggested changes.

src/wireviz/DataClasses.py
```diff
@@ -2,11 +2,48 @@
...
+
+        if self.fixedsize:
+            # If only one dimension is specified, compute the other
+            # because Graphviz requires both when fixedsize=True.
+            if self.height:
+                if not self.width:
+                    self.width = self.height * aspect_ratio(gv_dir.joinpath(self.src))
+            else:
+                if self.width:
+                    self.height = self.width / aspect_ratio(gv_dir.joinpath(self.src))
```
kvid, line 44: Perhaps add an `else` below this line:
```py7
else
    raise Exception('fixedsize=True and no dimension is specified!')
```

formatc1702: Let's include your suggestion, it's great that you are thinking of all these possible edge cases!

formatc1702:
One last thing, and this PR will finally be merged!
Please add PIL / pillow to both `requirements.txt` and `setup.py` (under `install_requires=[`).
Thanks again for this great addition to the project!

kvid:
> One last thing, and this PR will finally be merged!
> Please add PIL / pillow to both `requirements.txt` and `setup.py` (under `install_requires=[`).
> I think it's more elegant to have all dependencies installed, even if a user might not use images. Your error catching is still great, if for some reason PIL is uninstalled.. that way WireViz will fall back to the default 1:1 ratio.
> Thanks again for this great addition to the project!

I have implemented your request, but after testing it, I have some questions:

Why do you want the run-time dependencies both in `requirements.txt` and in `setup.py` when the latter seems to be enough?
- The first line of `requirements.txt` contains `.`, and that means, AFAIK, that `setup.py` is executed to find dependencies specified there, and then adding the same dependencies again from the remaining lines of `requirements.txt` will not change anything.
- The last line of `requirements.txt` contains `setuptools`, but if that is not already installed before running `pip install -r requirements.txt`, then it will fail running `setup.py` when processing the first line containing `.`.
- Moving `setuptools` above the `.` in `requirements.txt` will still fail if `setuptools` is not already installed, probably because all dependencies are processed before running any installation.
- Why do you really need `requirements.txt` in the repo at all? it does not seem to be used by the project installation instructions. Maybe the installation instructions should have an initial `pip install setuptools` to make sure that is installed before running `setup.py`?

formatc1702:
TBH, I am not too familiar with the processes behind `pip`, `setuptools`, etc.
I have seen a lot of projects using both, and [this post](https://stackoverflow.com/questions/43658870/requirements-txt-vs-setup-py) for example, seems to say that, while there is a way to reduce the duplicate references, if is not unreasonable to keep both files. That's why, for now, I'd like to go the safe route and keep both, too.
In the future, we could open a new issue/PR to optimize the setup process, if it is deemed necessary...

kvid:
> TBH, I am not too familiar with the processes behind `pip`, `setuptools`, etc.
> I have seen a lot of projects using both, and [this post](https://stackoverflow.com/questions/43658870/requirements-txt-vs-setup-py) for example, seems to say that, while there is a way to reduce the duplicate references, if is not unreasonable to keep both files. That's why, for now, I'd like to go the safe route and keep both, too.
> In the future, we could open a new issue/PR to optimize the setup process, if it is deemed necessary...

I read the same post yesterday that you link to, and I concluded there are different opinions about the reason behind `requirements.txt`:
 - Some argue that it's used for installing everything needed for module development without installing the module, but then it doesn't make sense to include `.` (that does install this module).
 - Some argue to include just `.` and nothing else to let `setup.py` do the installation job, but then why store and use the `requirements.txt` at all?
 - Some argue to store the output from `pip freeze` into `requirements.txt` as a documentation on the exact development environment, but in this project we cannot expect all developers to use the exact same set of versions.
 - In any case it doesn't make sense to include both `.` and `setuptools` in the same `requirements.txt`, as far as I can see.

I will update both files when that is your request, even though I don't like this double specification of the exact same information. A new issue is a good idea, and I will probably create that later today.

formatc1702:
Yes, probably there are two of these views in our source code right now, clashing.

I think the first argument, of removing `.` from `requirements.txt` is the most reasonable, so let's do that for now. Since I am unsure of any other ramifications of emptying/removing `requirements.txt` at this point, let's keep the duplicate dependency info for now and discuss in a new issue.
Thanks!

formatc1702:
Please let me know when you've added the dependencies (even if they're still duplicated) and implemented your suggestion that I [commented on](https://github.com/formatc1702/WireViz/pull/153/files#r502491356).

We've talked about this before, but I'd like to know your opinion, just to be sure we agree:
I will probably squash this entire PR into one commit, since it's mostly additions/few modifications of existing code and, while the process to get here was long, the final implementation is very straightforward, so not much information should be lost if squashed. But if you prefer preserving the history, it would not be a problem either. :) Just let me know.

src/wireviz/DataClasses.py
kvid: This was my suggested addition:
```diff
------------------------- src/wireviz/DataClasses.py --------------------------
index 3ad84d2..e73de9c 100644
@@ -42,6 +42,8 @@ class Image:
             else:
                 if self.width:
                     self.height = self.width / aspect_ratio(gv_dir.joinpath(self.src))
+                else:
+                    raise RuntimeError('fixedsize=True and no dimension is specified!')
 
 
 @dataclass
```
Without this addition, Graphviz does not raise an Exception, as I first thought, but only outputs a warning when this issue occurs (and still generates output files - probably ignoring the invalid attribute value):
```
Warning: fixed cell size with unspecified width or height
in label of node W1
Warning: fixed cell size with unspecified width or height
in label of node W1
```
With the addition, WireViz raises an exception and exit:
```
  File "./wireviz.py", line 257, in <module>
    main()
  File "./wireviz.py", line 253, in main
    parse(yaml_input, file_out=file_out)
  File "./wireviz.py", line 56, in parse
    harness.add_cable(name=key, **attribs)
  File "./../wireviz/Harness.py", line 30, in add_cable
    self.cables[name] = Cable(name, *args, **kwargs)
  File "<string>", line 21, in __init__
  File "./../wireviz/DataClasses.py", line 148, in __post_init__
    self.image = Image(**self.image)
  File "<string>", line 8, in __init__
  File "./../wireviz/DataClasses.py", line 46, in __post_init__
    raise RuntimeError('fixedsize=True and no dimension is specified!')
RuntimeError: fixedsize=True and no dimension is specified!
```
To provoke this issue, the YAML input must contain an image with `fixedsize: True` and no `width` nor `heigth` - which is probably not a simple mistake done by the average user. Are you still sure you prefer the exception before the Graphviz warning? If you prefer the exception, I would suggest to include the connector/cable _name_ (as the wireviz warning does), but that information is currently not available inside the Image dataclass.

kvid:
> Please let me know when you've added the PIL dependencies (even if they're still duplicated) and implemented your suggestion that I [commented on](https://github.com/formatc1702/WireViz/pull/153/files#r502491356).
> #172 can be resolved later once we reach a decision on how to deal with the duplication.
> 
> We've talked about this before, but I'd like to know your opinion, just to be sure we agree:
> I will probably squash this entire PR into one commit, since it's mostly additions/few modifications of existing code and, while the process to get here was long, the final implementation is very straightforward, so not much information should be lost if squashed. But if you prefer preserving the history, it would not be a problem either. :) Just let me know.

I have committed these changes locally, but awaits your opinion about adding the exception - see https://github.com/formatc1702/WireViz/pull/153#discussion_r503003474.

I agree #172 can be fully resolved later, but please read my https://github.com/formatc1702/WireViz/issues/172#issuecomment-706839829 about removing the conflict before releasing v0.2.

I also agree squashing this entire PR into one commit. Can you use the current PR title and description as commit subject and body, or should I add something more? Suggested commit message:
```
Add optional image to connectors and cables (#153)

This image is displayed as part of the connector/cable node in the diagram.

This solves issue #27, and is a continuation of PR #137 that was closed due to changes in the base branch.
```

src/wireviz/DataClasses.py
formatc1702: Hmm, in this case, the GraphViz warning is more helpful, and it clearly specifies which node is causing the problem. You have convinced me that the current implementation is fine, and as you said, this is rather unlikely to happen anyway. Thanks!

formatc1702:
I have replied to your comment regarding the exception (--> leave as is, let GraphViz issue a warning).
The quick solution to #172 seems to have settled on removing `.` from `requirements.txt`, so that is done as well.
The commit message is perfectly fine! :)

Then let's get rid of that `.` and close this PR. Thanks.

kvid:
> I have replied to your comment regarding the exception (--> leave as is, let GraphViz issue a warning).

OK, I have now pushed my commits without the exception.

> The quick solution to #172 seems to have settled on removing `.` from `requirements.txt`, so that is done as well.

I'm not sure I would say the solution "has settled" when only one contributor has responded and his comment is not very conclusive, but rather a bit vague... 😞 

> The commit message is perfectly fine! :)

Should I do the squash rebase to keep my authorship or will that be unchanged even if you do the squashing?

> 
> Then let's get rid of that `.` and close this PR. Thanks.

IMHO, removing a conflict in `requirements.txt` is out of scope for this PR. I don't like to hide such a change in an unrelated squash commit.

formatc1702:
> > The quick solution to #172 seems to have settled on removing `.` from `requirements.txt`, so that is done as well.
> 
> I'm not sure I would say the solution "has settled" when only one contributor has responded and his comment is not very conclusive, but rather a bit vague... 😞

True. But nothing is stopping us from adding `pillow` to both files right now, to close this PR, right?

> > The commit message is perfectly fine! :)
> 
> Should I do the squash rebase to keep my authorship or will that be unchanged even if you do the squashing?

AFAIK, authorship should be preserved. Let's try. If it does not work, I will roll back, and it will be a learning experience ;)

> > Then let's get rid of that `.` and close this PR. Thanks.
> 
> IMHO, removing a conflict in `requirements.txt` is out of scope for this PR. I don't like to hide such a change in an unrelated squash commit.

Fine, I agree.

kvid:
> > > The quick solution to #172 seems to have settled on removing `.` from `requirements.txt`, so that is done as well.
> > 
> > I'm not sure I would say the solution "has settled" when only one contributor has responded and his comment is not very conclusive, but rather a bit vague... 😞
> 
> True. But nothing is stopping us from adding `pillow` to both files right now, to close this PR, right?

No, I believe the current state of this branch can be used. I have checked the raw logs from [the latest 3.7 and 3.8 tests](https://github.com/formatc1702/WireViz/pull/153/checks), and did not find any warning outputs. That indicates importing from PIL went OK when `aspect_ratio()` was called for `ex08.yml`. Ideally, we should test installing from scratch at all supported platforms and Python versions, but I currently are not able to do that. What are the test plans before a release?

Since this PR is only a partial solution to #27, I assume you keep #27 open so the discussions about more advanced integration of images are not so easily forgotten.

> 
> > > The commit message is perfectly fine! :)
> > 

OK

> > 
> > Should I do the squash rebase to keep my authorship or will that be unchanged even if you do the squashing?
> 
> AFAIK, authorship should be preserved. Let's try. If it does not work, I will roll back, you squash, and it will be a learning experience ;)

OK

> 
> > > Then let's get rid of that `.` and close this PR. Thanks.
> > 
> > IMHO, removing a conflict in `requirements.txt` is out of scope for this PR. I don't like to hide such a change in an unrelated squash commit.
> 
> Fine, I agree. Let's keep things transparent and split it into a new PR.

Good. That was also one of the reasons behind the #172 issue.

formatc1702:
> > True. But nothing is stopping us from adding `pillow` to both files right now, to close this PR, right?
> 
> No, I believe the current state of this branch can be used. I have checked the raw logs from [the latest 3.7 and 3.8 tests](https://github.com/formatc1702/WireViz/pull/153/checks), and did not find any warning outputs. That indicates importing from PIL went OK when `aspect_ratio()` was called for `ex08.yml`. Ideally, we should test installing from scratch at all supported platforms and Python versions, but I currently are not able to do that. What are the test plans before a release?

Honestly, right now I am happy with seeing the passing builds that you linked to. There's no plan to do exhaustive testing at this point; I'd rather release first and wait for eventual bug reports... we are still at a very early stage and, unless we can somehow automate more exhaustive testing (new issue?), I will not worry too much about it.

While aiming for quality is a noble goal, I'd like to focus on moving things along quickly, the way we have been (at least before my holiday). I hope you understand.

formatc1702:
4782da47c9060304d61c947a1a6b6ab27aa53bb8 was squashed and merged, preserving your authorship. Thanks for this valuable addition to WireViz! 🎉

PR #154, opened 2020-08-15
Add .gitattributes for GitHub Linguist

formatc1702:
This is an attempt to
- avoid the `examples/` and `tutorial/` directories to be included in the project's language statistics
- mark any WireViz output files (TSV, CSV, GV, HTML, SVG, PNG) as generated files

Unfortunately, it seems that the statistics are only updated when merged into `master`, so maybe someone with more experience can confirm that my implemenation will actually have the intended effect.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,11 @@
+docs/* linguist-documentation
+
+examples/* -linguist-detectable
+tutorial/* -linguist-detectable
+
+**/*.bom.tsv linguist-generated
+**/*.bom.csv linguist-generated
+**/*.gv linguist-generated
+**/*.html linguist-generated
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,11 @@
+docs/* linguist-documentation
```
formatc1702, line 1: Coming in #111

.gitattributes
```diff
@@ -0,0 +1,11 @@
...
+
+examples/* -linguist-detectable
+tutorial/* -linguist-detectable
```
formatc1702, line 4: Does it make sene to move `examples/` and `tutorial` _into_ the `docs/` directory?

PR #161, opened 2020-08-28
Adding support for wire length units

zombielinux:
Will also be editing dataclasses.py to add the support as well as defaulting units to meters if unpopulated.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -170,7 +170,7 @@ class Harness:
                      f'{cable.wirecount}x' if cable.show_wirecount else None,
                      f'{cable.gauge} {cable.gauge_unit}{awg_fmt}' if cable.gauge else None,
                      '+ S' if cable.shield else None,
-                     f'{cable.length} m' if cable.length > 0 else None,
+                     f'{cable.length} {cable.lengthunit}' if cable.length > 0 else None,
                      cable.color, '<!-- colorbar -->' if cable.color else None],
                     '<!-- wire table -->',
                     [html_line_breaks(cable.notes)]]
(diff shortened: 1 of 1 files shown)


src/wireviz/Harness.py
```diff
@@ -170,7 +170,7 @@ def create_graph(self) -> Graph:
                      f'{cable.wirecount}x' if cable.show_wirecount else None,
                      f'{cable.gauge} {cable.gauge_unit}{awg_fmt}' if cable.gauge else None,
                      '+ S' if cable.shield else None,
-                     f'{cable.length} m' if cable.length > 0 else None,
+                     f'{cable.length} {cable.lengthunit}' if cable.length > 0 else None,
```
kvid, line 173: I can see an unfortunate difference in naming style between these two attribute pairs:
- `cable.gauge` and `cable.gauge_unit`
- `cable.length` and `cable.lengthunit`

Feel free to discuss naming conventions in a new issue, but I suggest we try to keep it consistent.

kvid (changes requested):
- Just a friendly advice: Read the hints about writing good commit messages: https://github.com/formatc1702/WireViz/blob/feature/syntax-description/docs/CONTRIBUTING.md#hints
- Please include all commits needed for this feature into your `patch-1` branch and re-push it, i.e. join (rebase or cherry-pick) the #162 commit(s) into this PR. The automatic PR checks currently fail in this PR because the `DataClasses.py` changes are missing.

src/wireviz/Harness.py
stevegt: added the underbar to length_unit in #171

stevegt:
merged with #162 and submitted as #171

formatc1702:
Closed as part of #196.

PR #163, opened 2020-09-01
Update the types of dataclass attributes according to usage

### LINKED ISSUE #156: [internal] Using the Optional[] keyword in dataclasses
Currently, some fields of the classes in `DataClasses.py` use the `Optional[]` keyword, and some don't.
```python
class Connector:
    ...
    color: Optional[str] = None
    show_name: bool = None
```

I am unsure if there is a difference in behavior between both; or if the ` = None` is needed at all.
Boolean values should probably have a default value of `False` anyway, instead of `None`.

Perhaps someone can chime in and suggest which variant is recommended; then I can create a PR to make the syntax more consistent. Thanks!

kvid:
Fixes #156

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -79,8 +79,8 @@ class Connector:
     pinlabels: List[Any] = field(default_factory=list)
     pins: List[Any] = field(default_factory=list)
     color: Optional[str] = None
-    show_name: bool = None
-    show_pincount: bool = None
+    show_name: Optional[bool] = None
+    show_pincount: Optional[bool] = None
     hide_disconnected_pins: bool = False
     autogenerate: bool = False
     loops: List[Any] = field(default_factory=list)
@@ -167,7 +167,7 @@ class Cable:
     length: float = 0
     color: Optional[str] = None
     wirecount: Optional[int] = None
-    shield: bool = False
+    shield: Union[bool, str] = False # False | True | color
     image: Optional[Image] = None
     notes: Optional[str] = None
     colors: List[Any] = field(default_factory=list)
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -1,15 +1,32 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
-from typing import Optional, List, Any, Union
+from typing import Optional, List, Tuple, Union
 from dataclasses import dataclass, field, InitVar
 from pathlib import Path
 from wireviz.wv_helper import int2tuple, aspect_ratio
 from wireviz import wv_colors
 
+
+# Each type alias have their legal values described in comments - validation might be implemented in the future
+PlainText = str # Text not containing HTML tags nor newlines
+Hypertext = str # Text possibly including HTML hyperlinks that are removed in all outputs except HTML output
+MultilineHypertext = str # Hypertext possibly also including newlines to break lines in diagram output
+Designator = PlainText # Case insensitive unique name of connector or cable
+
 # Literal type aliases below are commented to avoid requiring python 3.8
-ConnectorMultiplier = str  # = Literal['pincount', 'populated']
-CableMultiplier = str  # = Literal['wirecount', 'terminations', 'length', 'total_length']
+ConnectorMultiplier = PlainText # = Literal['pincount', 'populated']
+CableMultiplier = PlainText # = Literal['wirecount', 'terminations', 'length', 'total_length']
+ImageScale = PlainText # = Literal['false', 'true', 'width', 'height', 'both']
+Color = PlainText # Two-letter color name = Literal[wv_colors._color_hex.keys()]
+ColorScheme = PlainText # Color scheme name = Literal[wv_colors.COLOR_CODES.keys()]
+
+# Type combinations
+Colors = PlainText # One or more two-letter color names (Color) concatenated into one string
+Pin = Union[int, PlainText] # Pin identifier
+Wire = Union[int, PlainText] # Wire number or Literal['s'] for shield
+NoneOrMorePins = Union[Pin, Tuple[Pin, ...], None] # None, one, or a tuple of pins
+OneOrMoreWires = Union[Wire, Tuple[Wire, ...]] # One or a tuple of wires
 
 
 @dataclass
@@ -17,13 +34,13 @@ class Image:
     gv_dir: InitVar[Path] # Directory of .gv file injected as context during parsing
     # Attributes of the image object <img>:
     src: str
-    scale: Optional[str] = None  # false | true | width | height | both
+    scale: Optional[ImageScale] = None
     # Attributes of the image cell <td> containing the image:
     width: Optional[int] = None
     height: Optional[int] = None
     fixedsize: Optional[bool] = None
     # Contents of the text cell <td> just below the image cell:
-    caption: Optional[str] = None
+    caption: Optional[MultilineHypertext] = None
     # See also HTML doc at https://graphviz.org/doc/info/shapes.html#html
 
     def __post_init__(self, gv_dir):
@@ -47,13 +64,14 @@ class Image:
                 if self.width:
                     self.height = self.width / aspect_ratio(gv_dir.joinpath(self.src))
 
+
 @dataclass
 class AdditionalComponent:
-    type: str
-    subtype: Optional[str] = None
-    manufacturer: Optional[str] = None
-    mpn: Optional[str] = None
-    pn: Optional[str] = None
+    type: MultilineHypertext
+    subtype: Optional[MultilineHypertext] = None
+    manufacturer: Optional[MultilineHypertext] = None
+    mpn: Optional[MultilineHypertext] = None
+    pn: Optional[Hypertext] = None
     qty: float = 1
     unit: Optional[str] = None
     qty_multiplier: Union[ConnectorMultiplier, CableMultiplier, None] = None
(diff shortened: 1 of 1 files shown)


src/wireviz/DataClasses.py
```diff
@@ -90,7 +90,7 @@ class Cable:
     length: float = 0
     color: Optional[str] = None
     wirecount: Optional[int] = None
-    shield: bool = False
+    shield: Union[bool, str] = False # False | True | color
     notes: Optional[str] = None
     colors: List[Any] = field(default_factory=list)
     color_code: Optional[str] = None
```
kvid, line 96: Why have all attributes of the class `Connection` below `Any` as type? Can any of them have another type than `str` or `None`? If not, then I suggest `Optional[str]` instead of `Any` (or simply `str` for those who cannot be `None`).

kvid (commented):
A couple of my comments below is connected to the wrong code line because Github didn't allow commenting code lines that are not yet changed in this PR.

Are we using type annotations only for dataclasses, or should we use it for the rest of the code as well?

src/wireviz/DataClasses.py
```diff
@@ -1,32 +1,41 @@
...
+from typing import Optional, List, Union
 from dataclasses import dataclass, field
 from wireviz.wv_helper import int2tuple
 from wireviz import wv_colors
 
 
+Name = str # Case insensitive unique name of connector or cable
+Pin = Union[int, str] # Pin identifier
+Wire = Union[int, str] # Wire number or 's' for shield
+MLstr = str # Multi-line string where any newline is properly handled
```
kvid, line 13: If PR #164 is accepted, then we might need 3 different text types:
- Text = Text that might contain HTML hyperlinks that are removed except in HTML output.
- TextML = Text that might contain HTML hyperlinks that are removed except in HTML output, and newlines that are translated to `<br/>` in diagram output or to space otherwise.
- str = Plain text that doesn't contain any HTML tags nor newline.

formatc1702:
This is a very nice contribution.
I suggest changing `Name` to `Designator` since that is what has been used in the documentation and the BOM as well, and is a bit less generic.
While my initial reaction was skepticism at having too many custom datatypes, I can see the value in explicitly defining them so there are no surprises when HTML links are stripped, etc.
Now, my worry is just that `Text` and `TextML` are not self-explanatory enough (I know you added comments in their definitions)... Maybe something like `str_html` and `str_html_multiline`? I'm open for better suggestions

formatc1702:
According to your [suggestion](https://github.com/formatc1702/WireViz/issues/101#issuecomment-703197205), this PR should be the next one to be merged.
If you agree with my suggested naming changes, feel free to implement them and rebase onto `dev` :)
If you feel there is a need for additional discussion, fire away!

kvid:
I'm sorry for this late reply.

> This is a very nice contribution.

Thank's.

> I suggest changing `Name` to `Designator` since that is what has been used in the documentation and the BOM as well, and is a bit less generic.

I agree.

> While my initial reaction was skepticism at having too many custom datatypes, I can see the value in explicitly defining them so there are no surprises when HTML links are stripped, etc... because yes, #164 is a great new feature.
> Now, my worry is just that `Text` and `TextML` are not self-explanatory enough (I know you added comments in their definitions)... Maybe something like `str_html` and `str_html_multiline`? I'm open for better suggestions

[PEP 484](https://www.python.org/dev/peps/pep-0484/#type-aliases) recommends capitalizing alias names, since they represent user-defined types, which (like user-defined classes) are typically spelled that way.

I therefore suggest these alternatives:
- `PlainText = str` # Text not containing HTML tags nor newlines
- `LinkText = str` # Text possibly including HTML hyperlinks that are removed in all outputs except HTML output
- `MultilineLinkText = str` # LinkText possibly also including newlines to break lines in diagram output

formatc1702:
I'm sorry I keep nagging about the exact name of the user-defined types.

IMHO, `LinkText` creates the expectation that this is a text that always/usually is or includes a link.
However, as I mentioned in my [comment](https://github.com/formatc1702/WireViz/pull/168#issuecomment-716448650) in #168:
> TBH, having a separate `url` field will probably make the need for including URLs in the type attribute obsolete in most cases.

That's why I am uncomfortable pushing what I expect to be an edge case use for an attribute, into the very definition of that attribute.
My only counter-suggestion, however, might be `RichText` and `RichTextMultiline`, which on the other hand creates an expectation that it supports much more than just links.

Since I don't want this problem delaying integration much further, I will merge it as-is tomorrow unless we find a nicer alternative (or unless you want to change something else before, let me know!). We can always rename the type later if something better pops up.

kvid:
> I'm sorry I keep nagging about the exact name of the user-defined types.

Extra effort to find good identifier names should be an investment that hopefully pays off in the future by reducing the number of misunderstandings by users.

> 
> IMHO, `LinkText` creates the expectation that this is a text that always/usually is or includes a link.

I see your point. Is `TextThatMightContainLinks` or `TextWithOptionalLinks` any better?

> However, as I mentioned in my [comment](https://github.com/formatc1702/WireViz/pull/168#issuecomment-716448650) in #168:
> 
> > TBH, having a separate `url` field will probably make the need for including URLs in the type attribute obsolete in most cases.

I don't fully agree. Supporting link tags is way more flexible in the sense that the user can visualize what the link represents by selecting only a section of the text attribute as the clickable link text, and if needed, link to different URLs from different text sections.

> 
> That's why I am uncomfortable pushing what I expect to be an edge case use for an attribute, into the very definition of that attribute.

The whole point of type hints is to document what kind of values that are valid, so the type alias names should describe the differences somehow.

> My only counter-suggestion, however, might be `RichText`/`RichTextMultiline` or `String_HTML`, which on the other hand creates an expectation that it supports much more than just links.

A better alternative is perhaps `Hypertext` that is defined by [Wikipedia](https://en.wikipedia.org/wiki/Hypertext) as _text displayed on a computer display or other electronic devices with references (hyperlinks) to other text that the reader can immediately access._

If you like `RichText` better, it can be explained as text that might contain tags for the HTML output, but currently only supporting link tags, and add support for more tags later on. However, I believe the Rich Text term is more used to describe other formats and markup languages that HTML, e.g. RTF.

> 
> Another option is getting rid of the custom types, marking everything as `str` and adding a little comment in the dataclass mentioning the fact that a certain attribute has the custom Link/Linebreak behavior. Personally, I am leaning towards this option again after all...

If we are going to use type hints in more parts of the source code that DataClasses, then the different type aliases will get more useful to avoid the same comments at several locations.

> 
> Since I don't want this problem delaying integration much further, I will merge it as-is tomorrow unless we find a nicer alternative (or unless you want to change something else before, let me know!). We can always rename the type later if something better pops up.

True, but I'm willing to think a bit more before we decide.

formatc1702:
`Hypertext`! That's a good one!

It almost sounds a bit retro (in a good way), but per the definition you posted, it's pretty close to what we are trying to communicate through the type hint, without over-promising (like `RichText`).

I wouldn't feel bad including `Hypertext` and `HypertextMultiline` (whatever suits you best), thanks for the cool suggestion!

kvid:
> `Hypertext`! That's a good one!

Then we can agree on that one. 😃 

> 
> It almost sounds a bit retro (in a good way), but per the definition you posted, it's pretty close to what we are trying to communicate through the type hint, without over-promising (like `RichText`).

The term is from 1963 and has been used in different projects and predecessors of HTML/HTTP. The term is also the leading and major part of both HTML and HTTP.

> 
> I wouldn't feel bad including `Hypertext` and `HypertextMultiline`, thanks for the cool suggestion!

Personally, I feel `MultilineHypertext` is easier to read as it has the natural order of these words in an English sentence. Do you have strong feelings or arguments for the opposite order?

formatc1702:
> > It almost sounds a bit retro (in a good way), but per the definition you posted, it's pretty close to what we are trying to communicate through the type hint, without over-promising (like `RichText`).
> 
> The term is from 1963 and has been used in different projects and predecessors of HTML/HTTP. The term is also the leading and major part of both HTML and HTTP.

Of course it's still relevant; I just haven't heard anybody use the actual term "hypertext" in conversation or in writing for a veeery long time ;-)

> Personally, I feel `MultilineHypertext` is easier to read as it has the natural order of these words in an English sentence. Do you have strong feelings or arguments for the opposite order?

My main argument would be that, in certain situations, it makes sense to assign names starting from the general, then going into the specific. In this particular case, it's not really that important, so `MultilineHypertext` is fine too!

formatc1702:
Is this ready for merging from your perspective? I saw you force-pushed the latest changes, but I'm not sure.

I would like to try something out: Once you are done with your changes, please use the link in the top right of this PR to request a review from me. That lets me know you are finished with your changes, and if I approve the review, I can merge without having to ask again. Thanks!

kvid:
> Is this ready for merging from your perspective? I saw you force-pushed the latest changes, but I'm not sure.

Yes, now I think it should be ready. I squashed most commits into one, but kept the first commit separate to avoid hiding this separate issue in all the type alias changes afterwords.

> 
> I would like to try something out: Once you are done with your changes, please use the link in the top right of this PR to request a review from me. That lets me know you are finished with your changes, and if I approve the review, I can merge without having to ask again. Thanks!

I would like to test this feature, but I can't find the link you describe. I found [this old blog posting about it](https://github.blog/2016-12-07-introducing-review-requests/), but I cannot find the cog wheel icon described there. This how my right column looks like:
image

src/wireviz/DataClasses.py
```diff
@@ -47,13 +64,14 @@ def __post_init__(self, gv_dir):
...
-    type: str
-    subtype: Optional[str] = None
-    manufacturer: Optional[str] = None
-    mpn: Optional[str] = None
-    pn: Optional[str] = None
+    type: MultilineHypertext
+    subtype: Optional[MultilineHypertext] = None
+    manufacturer: Optional[MultilineHypertext] = None
+    mpn: Optional[MultilineHypertext] = None
+    pn: Optional[Hypertext] = None
```
formatc1702, line 74: Maybe the answer to my question is buried in the discussion on #115, but:

Why are `manufacturer` and `mpn` multiline-capable, and `pn` is not?

Personally, I'm not sure why any of them would need to be multiline, but at least it should be consistend, since all of them are passed through the `html_line_breaks()` function.

formatc1702:
According to the [GitHub documentation](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/requesting-a-pull-request-review):

> Pull request authors can't request reviews unless they are either a repository owner or collaborator with write access to the repository.

This doesn't make sense to me, but I guess it's how things work. So for now, the only option I see is for the PR author to post an explicit comment stating that the code is ready for [re-]review :/

src/wireviz/DataClasses.py
kvid: > Maybe the answer to my question is buried in the discussion on #115, but:

This feature is much older that #115.

> 
> Why are `manufacturer` and `mpn` multiline-capable, and `pn` is not?

[Blame](https://github.com/formatc1702/WireViz/blame/dev/src/wireviz/Harness.py) tells me the change was committed by you in https://github.com/formatc1702/WireViz/commit/102c7d611377433c93b6227a8160c83106d594bd as suggested by me in https://github.com/formatc1702/WireViz/pull/136#discussion_r463942757: _In case a user need a long manufacturer info, I suggest supporting line breaks to avoid a very wide node._ And since both `manufacturer` and `mpn` was placed in the same table cell, I suggested calling `html_line_breaks` on the whole cell contents.

> 
> Personally, I'm not sure why any of them would need to be multiline, but at least it should be consistend, since all of them are passed through the `html_line_breaks()` function.

Currently, (since the commit mentioned above that is included in v0.2)  `manufacturer` and `mpn` are passed through the `html_line_breaks()` function, but not `pn`, and that's why I used different type hints to reflect how it is currently implemented.

I understand your argument about consistency, but then we also need to change the implementation. I have an idea about moving `html_line_breaks()` in #168 that will solve your consistency request, but it's still WIP.

formatc1702: Thanks for digging through history for me :) 

OK, then let's keep it as-is to reflect the current implementation, and I'll wait for #168.

PR #164, opened 2020-09-09
Remove HTML links from the input attributes 

martonmiklos:
This change will remove HTML links from the input attributes when generating graphs with GraphViz (because it does not support this HTML tag).

This way these links could be used generated the HTML BoM table (basically without changing the input) where they could be useful.

```
# Example 7: Crossover Cable
connectors:
  X1:
    type: '<a href="https://www.digikey.com/product-detail/en/stewart-connector/SS-37000-002/380-1098-ND/1033365">Stewart Connector SS-37000-002</a>'
    subtype: male
    pinlabels: [DA+,DA-,DB+,DC+,DC-,DB-,DD+,DD-] # pincount is implicit in pinout
  X2:
    type: '<a href="https://www.digikey.com/product-detail/en/stewart-connector/SS-37000-002/380-1098-ND/1033365">Stewart Connector SS-37000-002</a>'
    subtype: male
    pinlabels: [DB+,DB-,DA+,DD+,DD-,DA-,DC+,DC-]

```
Results:
kép

BTW.  it might be useful to add some links to some examples just to make some attention on this feature.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -8,7 +8,7 @@ from wireviz.wv_colors import get_color_hex
 from wireviz.wv_helper import awg_equiv, mm2_equiv, tuplelist2tsv, \
     nested_html_table, flatten2d, index_if_list, html_line_breaks, \
     clean_whitespace, open_file_read, open_file_write, html_colorbar, \
-    html_image, html_caption, manufacturer_info_field, component_table_entry
+    html_image, html_caption, manufacturer_info_field, component_table_entry, remove_links
 from collections import Counter
 from typing import List, Union
 from pathlib import Path
@@ -92,8 +92,8 @@ class Harness:
 
             html = []
 
-            rows = [[connector.name if connector.show_name else None],
-                    [f'P/N: {connector.pn}' if connector.pn else None,
+            rows = [[remove_links(connector.name) if connector.show_name else None],
+                    [f'P/N: {remove_links(connector.pn)}' if connector.pn else None,
                      html_line_breaks(manufacturer_info_field(connector.manufacturer, connector.mpn))],
                     [html_line_breaks(connector.type),
                      html_line_breaks(connector.subtype),
@@ -162,8 +162,8 @@ class Harness:
                 elif cable.gauge_unit.upper() == 'AWG':
                     awg_fmt = f' ({mm2_equiv(cable.gauge)} mm\u00B2)'
 
-            rows = [[cable.name if cable.show_name else None],
-                    [f'P/N: {cable.pn}' if (cable.pn and not isinstance(cable.pn, list)) else None,
+            rows = [[remove_links(cable.name) if cable.show_name else None],
+                    [f'P/N: {remove_links(cable.pn)}' if (cable.pn and not isinstance(cable.pn, list)) else None,
                      html_line_breaks(manufacturer_info_field(
                         cable.manufacturer if not isinstance(cable.manufacturer, list) else None,
                         cable.mpn if not isinstance(cable.mpn, list) else None))],
@@ -205,7 +205,7 @@ class Harness:
                     # create a list of wire parameters
                     wireidentification = []
                     if isinstance(cable.pn, list):
-                        wireidentification.append(f'P/N: {cable.pn[i - 1]}')
+                        wireidentification.append(f'P/N: {remove_links(cable.pn[i - 1])}')
                     manufacturer_info = manufacturer_info_field(
                         cable.manufacturer[i - 1] if isinstance(cable.manufacturer, list) else None,
                         cable.mpn[i - 1] if isinstance(cable.mpn, list) else None)
--- a/src/wireviz/wv_helper.py
+++ b/src/wireviz/wv_helper.py
@@ -3,6 +3,7 @@
 
 from wireviz import wv_colors
 from typing import List
+import re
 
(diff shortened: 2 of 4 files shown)


src/wireviz/wv_helper.py
```diff
@@ -113,6 +114,13 @@ def tuplelist2tsv(inp, header=None):
...
 
+def remove_links(inp):
+    if (inp is None):
+        return None
+    ret = inp.replace("</a>", "")
+    ret = re.sub(r"<a[^>]*>", "", ret)
+    return ret
+
 def html_line_breaks(inp):
     return inp.replace('\n', '<br />') if isinstance(inp, str) else inp
```
kvid, line 125: I suggest this code instead:
```py
return remove_links(inp).replace('\n', '<br />') if isinstance(inp, str) else inp
```
Then most calls to `remove_links()` in `Harness.py` can be removed.

src/wireviz/wv_helper.py
```diff
@@ -113,6 +114,13 @@ def tuplelist2tsv(inp, header=None):
 def index_if_list(value, index):
     return value[index] if isinstance(value, list) else value
```
kvid, line 116: I suggest replacing the second last code line of `tuplelist2tsv()` above with:
```py
output = output + '\t'.join(str(remove_links(item)) for item in row) + '\n'
```
Otherwise, the links will be included in the `.bom.tsv` output file.

kvid (changes requested):
@martonmiklos Thank you for contributing. I'm just another contributor, like you. @formatc1702 is the owner that approves and merges PRs into `dev` when he thinks they are ready. He is currently on vacation, and I decided to give you a few advices that I believe will help you getting this PR accepted, but remember that @formatc1702 might not share all my opinions.

- Just a friendly advice: Read the hints about writing good commit messages linked from this current draft written by @formatc1702 : https://github.com/formatc1702/WireViz/blob/feature/syntax-description/docs/CONTRIBUTING.md
- A couple of my code comments below are not attached to the correct code line because github doesn't allow attaching comments to code lines too far away from code changes in this PR.

src/wireviz/Harness.py
```diff
@@ -93,12 +93,12 @@ def create_graph(self) -> Graph:
             rows = [[connector.name if connector.show_name else None],
                     [f'P/N: {connector.pn}' if connector.pn else None,
```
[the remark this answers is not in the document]
martonmiklos: Fixed

martonmiklos:
Many thanks for your valuable feedback @kvid !
I agree and fixed with all of your comments.

kvid (changes requested):
Thank you again that you accepted my suggestions. I'm sorry for nagging on you again, but I have some new suggestions today as well:
- I believe @formatc1702 prefer that you avoid committing changes to generated files in PRs to reduce merging conflicts. If you use a Linux-shell or something similar, then you can execute this command to restore the generated files from the `dev` branch:
  `git checkout dev {examples/ex05,tutorial/tutorial08}.{gv,bom.tsv,png,svg,html}`
- Personally, I always execute `python build_examples.py restore` before committing to avoid including changes to generated files after testing with `python build_examples.py`.
- I suggest you also demonstrate hyperlinks from something in `additional_bom_items` in tutorial 8 to show that is supported too. If you find a real label product, adapt the entry for that, or find a totally different entry with hyperlinks that you can insert in addition to the labels entry. If it makes sense, consider also demonstrating two links for the same entry, e.g. both an `mpn` link and either a `manufacturer` main page or some generic link in (part of) the `description`.
- I do like that you improved your commit message, and I don't like to criticize that again, but have you noticed that github is truncating your subject line with an ellipsis? That is a hint telling you that the subject line is longer than recommended. It might be hard to find a short subject line sometimes, but e.g. "**Remove hyperlinks in texts except in the HTML BOM**" is a shorter alternative (less than 50 characters that is the soft limit). I believe [BOM](https://en.wikipedia.org/wiki/Bill_of_materials) is more common than BoM, but you decide what to write.

martonmiklos:
Many thanks for the review again @kvid !

> I believe @formatc1702 prefer that you avoid committing changes to generated files in PRs to reduce merging conflicts - see #17 (comment). 

Maybe we should add this info to the contribution guidelines?

kvid (approved):
> Many thanks for the review again @kvid !

Thank's again for accepting my latest suggestions. In my opinion, this PR is now ready for merge into `dev`, but @formatc1702 might disagree.

> 
> > I believe @formatc1702 prefer that you avoid committing changes to generated files in PRs to reduce merging conflicts - see #17 (comment).
> 
> Maybe we should add this info to the contribution guidelines?

I fully agree with you, and included this in my https://github.com/formatc1702/WireViz/pull/111#discussion_r487465380.

formatc1702:
As per @kvid's [suggestion](https://github.com/formatc1702/WireViz/issues/101#issuecomment-703197205), this should be the next PR to be merged into `dev`.

Therefore, please rebase onto `dev` to get rid of any merge conflicts, and have a look if anything needs changing.

I will squash this PR into a single commit since the actual code change is very minor, so no need to interactively rebase/restructure the commit history, IMHO.

Thanks!

martonmiklos:
Hi @formatc1702 

I have just rebased and squashed my changes to my branch to the dev.

formatc1702:
Thank you for your contribution! :)

PR #173, opened 2020-10-10
Change command line options

kvid:
Solves issue #167

Issue #97, opened 2020-07-17
[bug] Improper XML and DOCTYPE Declarations


kvid:
The HTML output violates these rules:
- [If the XML declaration is included, it must be situated at the first position of the first line in the XML document.](https://xmlwriter.net/xml_guide/xml_declaration.shtml)
- [The document type declaration must be placed between the XML declaration and the first element (root element) in the document.](https://xmlwriter.net/xml_guide/doctype_declaration.shtml)

The root cause is that `Harness.output()` start HTML output with [`file.write('<html><body style="font-family:Arial">')`](https://github.com/formatc1702/WireViz/blob/861380ddd64082a6b478f3d8acf8680ae3b67356/src/wireviz/Harness.py#L307) (i.e. no leading XML or DOCTYPE declarations) and then later embed the SVG diagram into the HTML output without hiding (remove or comment) the leading XML and DOCTYPE declarations from the SVG file.

If we are including XML and/or DOCTYPE declarations, then we should comply to their rules. Related issues I have not investigated:
- Does all the HTML generated by `Harness.output()` comply to identical XML or DOCTYPE declarations as the SVG file?
- Can a proper XML declaration with `encoding="UTF-8"` be of any help in issue https://github.com/formatc1702/WireViz/issues/52 ?

aakatz3:
The svg doctype is generated by graphviz itself. The best way to deal with this would be a complete refactor of HTML generation, using a library. I'm not amazing at python, but I found a couple that can be tried.

kvid:
> The svg doctype is generated by graphviz itself.

I know that, but it should be easy to filter out the leading declarations before embedding the rest of the SVG file.

> The best way to deal with this would be a complete refactor of HTML generation, using a library. I'm not amazing at python, but I found a couple that can be tried.

I cannot see how a library will solve this issue. It might be a good idea for other reasons, but that belong in a separate issue. If we want to add proper leading declarations, then it is easy to include in the first `file.write()`. IMHO, "a complete refactor of HTML generation" is overkill in this case.

The question is: Do we want any of these declarations at all?

aakatz3:
My thought was that if we use the xml.dom.minidom library, we can easily strip out or modify tags as necessary from the Graphviz SVG. Similarly, if we use that library, or a similar one for HTML, we can then just pass an object instead of repeatedly reading and writing files. 

[...]

PR #177, opened 2020-10-14
Add version number to output files with meta info

kvid:
This is a follow-up to the version number part of #173 to also tag the `.gv` and `.html` output files with version number.
The last few HTML validation errors remaining from #97 was also removed in the process.


src/wireviz/Harness.py
```diff
@@ -296,7 +296,11 @@ def output(self, filename: (str, Path), view: bool = False, cleanup: bool = True
         # HTML output
         with open_file_write(f'{filename}.html') as file:
             file.write('<!DOCTYPE html>\n')
-            file.write('<html><head><meta charset="UTF-8"></head><body style="font-family:Arial">')
+            file.write('<html lang="en"><head>\n')
+            file.write(' <meta charset="UTF-8">\n')
+            file.write(f' <meta name="generator" content="wireviz {__version__}" />\n')
```
kvid, line 301: Should we also add the Wireviz URL? How?

formatc1702:
This is a good idea and I'm willing to include it in v0.2!

src/wireviz/Harness.py
formatc1702: I am not too familiar with these `<meta>` tags... perhaps a HTML comment after this line is enough for the URL?

kvid: I didn't find any suitable [`<link>` relation values](https://www.w3schools.com/tags/att_link_rel.asp), [`<meta>` name values](https://www.w3schools.com/tags/att_meta_name.asp), [`<meta>` attributes](https://www.w3schools.com/tags/tag_meta.asp) or [global attributes](https://www.w3schools.com/tags/ref_standardattributes.asp), and was thinking just appending the URL after the version number, but I've never seen that used by others, so your suggestion is probably better.

Is it OK to keep the name as `WireViz` in the `.gv` output and `wireviz` most other places, or do you consider it unfortunate?

formatc1702: If it's meant to be human-readable (such as here), `WireViz` looks better IMHO. `wireviz` should be reserved for CLI commands etc.

[This site](http://www.meta-tags.de/meta_name_generator) even includes a copyright notice inside the `<meta>` tag... so it wouldn't really matter if the URL was in there too, after the version number.

`<meta name="generator" content="WireViz {__version__} - https://github.com/formatc1702/WireViz/">`

kvid: I edited my previous comment with a new suggestion before I discovered your last comment, but perhaps your last suggestion is a cleaner solution?

formatc1702: I think one line to include version and URL should be enough :) Thanks for looking into this.

kvid:
I'm sorry that the two last commits are slightly off topic, but the reasoning is:
- It didn't feel right to define the application name and URL once again when writing the meta tag (ref [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)), and with your https://github.com/formatc1702/WireViz/pull/177#discussion_r505555991 in mind, I searched for the application name in the code and found it written a lot of places - written 3 different ways: WireViz, Wireviz, and wireviz. Therefore, I moved the definition to a common file and imported these constants where needed.
- When validating the HTML meta tag, I first added a title tag to remove one of the errors, and finally I found a minor change of the HTML BOM that removed all the remaining errors. Do you want me to move this last commit to a separate PR, or should I just expand the scope slightly in the PR description?

formatc1702:
> I'm sorry that the two last commits are slightly off topic, but the reasoning is:
> 
> * It didn't feel right to define the application name and URL once again when writing the meta tag (ref [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)), and with your [#177 (comment)](https://github.com/formatc1702/WireViz/pull/177#discussion_r505555991) in mind, I searched for the application name in the code and found it written a lot of places - written 3 different ways: WireViz, Wireviz, and wireviz. Therefore, I moved the definition to a common file and imported these constants where needed.

I appreciate the attention to detail, and I agree having a standardized way to write the app's name (WireViz) is good.
However, defining this as a constant and then referencing it a couple of times, and especially piecing together the URL from it (`URL = 'https://github.com/formatc1702/' + APP_NAME`), is overdoing it and does not contribute to simplifying the code...
TBH I'd rather just correct the instances where the name/URL occurs, and make sure future PRs stick to the convention... perhaps adding a note to the contribution guidelines.
If the name/URL should ever change, a project-wide search&replace will be easy enough, I don't mind having it hardcoded in 2-3 places.

>     * When validating the HTML meta tag, I first added a title tag to remove one of the errors, and finally I found a minor change of the HTML BOM that removed all the remaining errors. Do you want me to move this last commit to a separate PR, or should I just expand the scope slightly in the PR description?

Hehe, I doubt this would have bothered anyone, and I doubt that the `ALIGN` attribute is ever going to truly stop working, but I am OK with fixing the validation error in this way. Up to you to create a new PR or not... I am more concerned with the commit history (which would end up looking pretty much the same either way). If you want to be extra clean about it, I guess a PR is the way to go.

src/wireviz/__init__.py
```diff
@@ -1,3 +1,7 @@
 # Please don't import anything in this file to avoid issues when it is imported in setup.py
 
 __version__ = '0.1.1'
+
+APP_NAME = 'WireViz'
+CMD_NAME = APP_NAME.lower()
```
formatc1702, line 6: `CMD_NAME = 'wireviz'` is shorter and easier to read, please change.

src/wireviz/__init__.py
```diff
@@ -1,3 +1,7 @@
...
+URL = 'https://github.com/formatc1702/' + APP_NAME
```
formatc1702, line 7: `URL = 'https://github.com/formatc1702/WireViz'` is fine, building up a URL using a constant feels like overkill.

formatc1702: Maybe rename to `APP_URL` to keep it consistent with `APP_NAME`

formatc1702:
Looks good, couldn't find any remaining references that would need fixing.

Much appreciated!

Issue #185, opened 2020-10-22
Create a "splice" type


MSBGit:
Sometimes I have one wire harness that is spliced to another wire harness (usually inside an enclosure for example). The splice is just wires soldered together and then heatshrinked, there is no physical connector at the splice. It would be useful to have a _splice_ type which functions like a connector but doesn't have pins or other connector attributes and would be rendered as one or more wires joining one or more other wires, but not in any type of "housing". Today, I create a _splice_ connector and use that to do what I want but it's confusing when someone looks at my wiring diagram and sees what looks like a connector but doesn't find a physical connector anywhere in the system (even though I comment the `yml` file when I define such a splice).
![splice type](https://user-images.githubusercontent.com/5497625/96819076-aad64e80-13d7-11eb-943e-b1fbaf0c0c4e.png)

formatc1702:
Have you considered the `style: simple` and `autogenerate: true` attributes?

Would the following example work for you?

```yaml
connectors:
  S: &splice
    style: simple
    autogenerate: true
    type: splice
  S1:
    <<: *splice
    autogenerate: false
    show_name: false

cables:
  W1:
    wirecount: 3
    color_code: DIN
  W2:
    wirecount: 4
    color_code: IEC

connections:
  -
    - W1: [1-3]
    - [S, S, S1]  # wires 1+2 use autogenerated splice connectors
                  # wire 3 uses splice connector with user-specified designator
    - W2: [1-3]
  -
    - S1          # reuse S1 for the three-way splice
    - W2: 4
```

![splice](https://user-images.githubusercontent.com/2447251/96830533-dab73d80-143b-11eb-8b73-356d940c3847.png)

See #184 for ongoing work on improving auto-generated connectors.

formatc1702:
@MSBGit 
if my suggestion helps you, and you feel #184 might make this kind of work easier in the future, please close this issue.
Otherwise, let us know more details about your problem, or perhaps a sketch of how you would wish your kind of harness would be rendered.

Issue #184, opened 2020-10-21
[feature] Improved auto-generated connectors and cables


formatc1702:
From the [syntax description](https://github.com/formatc1702/WireViz/blob/master/docs/syntax.md):

> If `autogenerate: true` is set, the connector will _not_ be generated at first. When defining the `connections` section (see below), every time the connector is mentioned, a new instance with an auto-incremented designator is generated and attached.
> 
> Since the auto-incremented and auto-assigned designator is not known to the user, one instance of the connector can not be referenced again outside the point of creation. The `autogenerate: true` option is therefore only useful for terminals with only one wire attached, or splices with exactly one wire going in, and one wire going out. If more wires are to be attached (e.g. for a three-way splice, or a crimp where multiple wires are joined), a separate connector with `autogenerate: false` and a user-defined, unique designator needs to be used.

An idea to remove the restriction of not knowing the designator of auto-generated connectors and not being able to reference them at a later point, would be to add a **new syntax feature** to a**llow the user to assign designators to auto-generated connectors**.

Proposed syntax:
```yaml
X.X1
```
[Edit 2020-10-22: changed from `->` to `.` to distinguish from arrows proposed in #134]

This would generate a new instance of component `X` and assign it the `X1` designator.
This is just an idea, an other possible alternatives come to mind:
- `X.X1`
- `X->X1` (possible confusion with arrow notation discussed in #134)
- `X:X1` (seems to be legal YAML, even when used as dict key `X:X1: [1-4]` but probably risky/confusing)
- `X$X1`
- `X1(X)`
- `X1@X`
- and many more
The following example takes this a step further, and proposes **auto-generating cables** as well.

Example for a Y-splice, reusing connectors, ferrules and cables through auto-generation:
```yaml
connectors:
  X:
    autogenerate: true
    pincount: 4

  F:
    autogenerate: true
    style: simple
    type: ferrule

cables:
  C:
    autogenerate: true
    wirecount: 4
    color_code: DIN

connections:
  -
    - X.X2: [1-4]
    - C.C1: [1-4]
    - [F.F1, F.F2, F.F3, F.F4]  # generate new instances of F and assign designators
    - C.C2: [1-4]
    - X.X2: [1-4]
  -
    - [F1, F2, F3, F4]  # use previously assigned designators
    - C.C3: [1-4]
    - X.X3: [1-4]
```
[...]

Issue #142, opened 2020-07-31
[feature] Enable including components in the harness


kvid:
Sometimes, a few simple components are integrated into the harness, like e.g. a LED indicator or switch mounted at a connector housing, a resistor at the end of a bus line, diodes for protecting against reversed polarity, etc. I suggest adding such a component as a connector, except that it can be marked as a Component instead of a Connector in the BOM entries.

In addition, it would be nice to have the option to connect the pins of such a component directly to a splice (simple connector) or connector pin without cable/wires between when that is how to do it in real life. It might be implemented similar to the mockup in https://github.com/formatc1702/WireViz/issues/134#issuecomment-665877504 where ferrules (simple connectors) are connected directly to another connector.

Are there other use cases where something like this might be useful?

formatc1702:
This sounds extremely useful, but will require some rewriting of the input checking logic, since so far it has been a strict CABLE-CONNECTOR-CABLE-CONNECTOR-.... sequence.

It's a good task to tackle for v0.3!

kvid:
> This sounds extremely useful, but will require some rewriting of the input checking logic, since so far it has been a strict CABLE-CONNECTOR-CABLE-CONNECTOR-.... sequence.

I have suggested a possible input syntax in https://github.com/formatc1702/WireViz/issues/134#issuecomment-671000507 that doesn't break the alternating connector-cable-sequence, except accepting a direct connection as an alternative to a cable in the sequence , and that also can support mating wire connectors within the harness, as in #148.

formatc1702:
Naïve question:
Is there a way to unify connectors and these components somehow in the YAML?
I understand that currently, the BOM entry auto-appends the "Connector" prefix, which is unwanted in this case, but easily fixed.

Alternatives:
- Rename `connectors` to `components`, and adding a `category: connector` parameter, which would not necessarily have to be rendered inside the node, but could be prepended in the BOM.
  - The problem is that, technically, cables and wires are components too, so the semantic distinction kind of falls apart.
  - Most components are likely to be connectors in the majority of use cases, so this adds a lot of unnecessary lines to the YAML if you need to specify this every time.
- Keep connectors and components separate, just like connectors and cables are separate.
  - BOM generation might get uglier, since now one more dictionary need to be parsed...
  - As mentioned above, a better semantic distinction would be helpful. 
- Simply omit the "Connector" in the BOM entry, and treat these components as connectors.
[...]

Issue #134, opened 2020-07-29
[feature] Add ferrule before connector


stmaxted:
It would be nice to be able to specify a wire requires a ferrule before being put into a connector. I'm think of the case where someone wants to do something like:
![IMG_20200729_160207](https://user-images.githubusercontent.com/17870845/88755337-0c11cb80-d1b5-11ea-9afe-fd1ad271de78.jpg)

kvid:
Maybe PR #115 could be a solution for this request?

formatc1702:
My gut feeling also is to specify the ferrules as additional BOM items (as per #115)...

otherwise, you would need to define a connection between two connectors directly, with no wire inbetween, which (at least currently) is not possible _by design_. I'm not saying it's impossible, though... the connection from ferrule to the actual connector could be graphically represented as a thin, dashed line, for example, or even an arrow, to make it clear that it's not a physical wire linking the two.

🤔

stmaxted:
Adding additional lines to the BOM would work nicely. 
However the mockup looks very similar to what I was thinking of.

kvid:
> otherwise, you would need to define a connection between two connectors directly, with no wire inbetween, which (at least currently) is not possible _by design_.
> I'm not saying it's impossible, though... the connection from ferrule to the actual connector could be graphically represented as a thin, dashed line, for example, or even an arrow, to make it clear that it's not a physical wire linking the two.

One way this could be specified, is to use `<-`, `--`, or `->` instead of a cable/wire between two connectors in a `connections` entry to indicate a direct connection without wires between, and the arrow direction can be used to indicate male to female connection like in the mockup above with a possible input YAML like this:
```yaml
connectors:
  X1:
    pincount: 4
    type: Screw terminal connector
    color: GN
  F1:
    style: simple
    autogenerate: true
    type: Ferrule
    color: GY

cables:
  W1:
    gauge: 0.25 mm2
    length: 1
    color: BK
    colors: [BK, WH, BU, BN]

connections:
  -
    - X1: [1-4]
    - <-
    - F1
    - W1: [1-4]
```
Such an implementation would also fix #148 to enable a harness that is split with mating connectors, and would help implementing #142 as well.

formatc1702:
I was thinking of having a separate section in the YAML file.
`connections` could remain unchanged, for strict connector-cable-connector-[...] sequences.
[...]

PR #186, opened 2020-10-22
Implement mates between connectors, better autogeneration

formatc1702:
This is a sandbox to try an implementation of #134 and #184 (with possible effects for #142 and #185).
Since both involve changing the WireViz syntax, I figured it might make sense to tackle both at the same time, to make sure there are no conflicts between the two syntax additions.

- Discussion regarding the new syntax for mating connectors happens in  #134.
- Discussion regarding the new syntax for auto-generated components happens in #184.

This PR will try to reflect the results in those discussions.

The implementation will work towards being able to parse the included `test/test1.yml` and produce the expected result. This test file will try to reflect and test any related new features suggested in the discussions.

No detailed code review is necessary at the moment, consider this a work in progress;


formatc1702:
The current state is able to parse all current examples/tutorials/demos (only minor changes to the source file necessary), as well as all test files, inspired by the discussions in the issues mentioned above. See `tests/` directory for sources.

<details>
 <summary>See renderings</summary>

## Test 1

test1

## Test 2

Test 2test2

## Test 3

test3

## Test 4

test4

## Test 9

test9


</details>

formatc1702:
Even though I have not worked on this feature for a while, I have been using it in practice and it has proven quite robust.

I would like to request a review of the code, and I am open for ideas on how to solve the [open TODO](https://github.com/formatc1702/WireViz/pull/186/files#diff-4868ecda012df03682c97a37b93917a698aeec9adac351873f6d78cf2afc908cR132-R138)...

@kvid how do you feel about this?

formatc1702:
Thanks @awb1015 for the feedback!

Usually I've waited for @kvid's approval before merging into `dev`, since he's quite good at catching edge cases, improving the code quality and providing generally valuable feedback. But as you can see by my previous comment, this might be a good PR to try out a new merging strategy, with post-hoc review and improvement.
I realize this would be an unilateral decision by me, and might be seen as hypocritical since kvid and others have had to wait for months for me to react to PRs and issues in the past, so I have been shy about this approach so far.

formatc1702:
Yes please, a setof files for a minimal working example will be really helpful.

formatc1702:
Please try the updated code; your bug should be fixed in 02a800abef3f35bdc4f8bbcafd9f50309a7cbe17.

Root cause:
- #229 changed the way connections between components are handled by internally assigning unique IDs to the "ports" where wires (and now arrows) attach to connectors.
- This mapping happened inside the `Harness.connect()` function, which is called when attaching wires.
  Arrows between pins use `Harness.add_mate_pin()` instead, which did not implement the mapping
- In addition, `Harness.create_graph()` had to be adjusted to add `+1` to convert the [0-indexed] internal IDs to 1-indexed port IDs.
  - The decision to use 1-indexed ports in GraphViz was taken so that port IDs would match the typical use case with connector pin numberings starting at 1
  - This is also the reason why the bug only appeared with your custom pin numbering.

kvid (changes requested):
Sorry about this long period without any reply. These changes looks good. I only have a couple of suggestions.

formatc1702:
All comments have been addressed inside https://github.com/formatc1702/WireViz/pull/251.

SeanDS:
This looks promising for my use case. Would it be possible to group together the two sides of the mate in one box, instead of those arrows? I'm thinking a box with numbering on both left and right sides and labels between indicating the name of the pin pair. As I noted in #295 there are certain connectors where pin numberings change on the same conductor, so these numbers would ideally not be fixed.

martinrieder:
It seems that the user-configurable autogeneration character from https://github.com/wireviz/WireViz/commit/45bcc1dd458a953ba1b2d6fe5ccd5ae866135661 became an undocumented feature.

kvid:
@martinrieder wrote:
> It seems that the user-configurable autogeneration character from [45bcc1d](https://github.com/wireviz/WireViz/commit/45bcc1dd458a953ba1b2d6fe5ccd5ae866135661) became an undocumented feature.

You are right, it's not mentioned in the docs, but I don't know if it might be due to lack of testing, unstability, just forgotten, or something else. Maybe @formatc1702 remembers any reason for omitting it from the docs?

formatc1702:
Thanks. I have added a line in the syntax description (in the `dev` branch) for now so it is not forgotten, see 590504130c5e878c51df571974ee31e6f01b3a53.

It's a good idea to do some testing as suggested; then the documentation can be expanded accordingly.

martinrieder:
@kvid wrote:
> Edit: If you @martinrieder have the time, maybe you could test this optional feature with a few different values, and verify the intended effect? Does it only work with a single character, or also with longer strings? Do you observe any unfortunate side-effects with certain characters/strings? Please store a few YAML input test cases for future regression testing.

I can do that. I guess that quotation characters could be problematic in this case. For sure, there should be a check to not accept an empty string here.

Issue #60, opened 2020-07-04
[feature] Only output the files that user requests


formatc1702:
Currently, running WireViz generates multiple output files from the YAML input by default:
- `.png` raster image
- `.svg` vector image
- `.gv` GraphViz code
- `.tsv` BOM
- `.html` page with embedded diagram and BOM (feature to be expanded, see # 32)

Especially the GraphViz code is probably not of much interest to the end user. [The `output()` function in Harness.py](https://github.com/formatc1702/WireViz/blob/dev/src/wireviz/Harness.py#L267) should be updated to only actually generate the files the user needs.

A good default behavior should be set, ideally just with one file type to avoid cluttering the file system.

aakatz3:
I had looked into this before, and there are a couple of different ways this can be done, from refactoring the outputs to simple if-else statements. The Graphviz DOT file can be useful as it can be converted into something like a Visio document, but (and I have to check the code,) I think its generated due to how Graphviz processes the DOT file to generate conversions. Since it externally calls the dot executable, and the dot executable needs to read from disk (at least by default), I think we would have to just delete the file if the user does not request it, unless Graphviz and the python library for it support sending the info via named pipes.

formatc1702:
Right now the .gv file is [explicitly saved](https://github.com/formatc1702/WireViz/blob/dev/src/wireviz/Harness.py#L297) so that's easy to fix.

I'd say a couple of `if` statements, the possibility to specify what to output when running from command line:
`$ python wireviz.py inputfile.yml -output png csv html` or similar
and setting good default behavior (my vote: generate PNG only so as not to generate clutter).

slightlynybbled:
If you are looking for a great library, check out [click](https://click.palletsprojects.com/en/7.x/).  Pure python and easy to maintain.  No chained `if` statements.  Have successfully used this on several internal CLI packages.

formatc1702:
That looks quite interesting.

@slightlynybbled would you be interested in taking over this issue and rewriting the [current CLI handling](https://github.com/formatc1702/WireViz/blob/dev/src/wireviz/wireviz.py#L210L248)?

Currently, the `--generate-bom` an `--output-file` flags are unused.

The usecases I envision:
`wireviz inputfile.yml` <- default behavior: output only PNG
`wireviz inputfile.yml --prepend library.yml` <- to include e.g. a `library` of templates/parts to reference in the `inputfile`
[...]

Issue #188, opened 2020-10-24
[feature] Ability to include Base64-encoded images inside YAML file


formatc1702:
Instead of linking to an external file, users should alternatively be able to include images as base64-encoded strings within the YAML file itself, to avoid the need for additional files.

This is not a high priority, and might be a niche use-case, but it could be an interesting addition at some point.

PR #189, opened 2020-10-24
Embed images in SVG output

formatc1702:
[Update 2020-11-16]

This PR has evolved from simply fixing how relative image paths are resolved, to embedding any images in the SVG file by default, eliminating the original issue.

---

Previously, paths to images were resolved relatively to the specified _output_ file instead of the _input_ file.

This produces errors when calling WireViz with the `-o` flag, creating output files in a different location from the input.

This PR [partially, see comment below] fixes the problem.

Additionally, it eliminates the need for passing a `gv_dir` variable to the `Image` class, resolving paths beforehand and passing the absolute path to the `Image` class.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -20,6 +20,7 @@ from wireviz.DataClasses import (
     Tweak,
     Side,
 )
+from wireviz.svgembed import embed_svg_images_file
 from wireviz.wv_bom import (
     HEADER_MPN,
     HEADER_PN,
@@ -646,13 +647,9 @@ class Harness:
 
     @property
     def svg(self):
-        from io import BytesIO
-
         graph = self.graph
-        data = BytesIO()
-        data.write(graph.pipe(format="svg"))
-        data.seek(0)
-        return data.read()
+        return embed_svg_images(graph.pipe(format="svg").decode("utf-8"), Path.cwd())
+
 
     def output(
         self,
@@ -671,9 +668,14 @@ class Harness:
             if f in ("png", "svg", "html"):
                 if f == "html":  # if HTML format is specified,
                     f = "svg"  # generate SVG for embedding into HTML
+                # SVG file will be renamed/deleted later
+                _filename = f"{filename}.tmp" if f == "svg" else filename
                 # TODO: prevent rendering SVG twice when both SVG and HTML are specified
                 graph.format = f
-                graph.render(filename=filename, view=view, cleanup=cleanup)
+                graph.render(filename=_filename, view=view, cleanup=cleanup)
+        # embed images into SVG output
+        if "svg" in fmt or "html" in fmt:
+            embed_svg_images_file(f"{filename}.tmp.svg")
         # GraphViz output
         if "gv" in fmt:
             graph.save(filename=f"{filename}.gv")
@@ -692,8 +694,11 @@ class Harness:
             # TODO: implement PDF output
             print("PDF output is not yet supported")
         # delete SVG if not needed
-        if "html" in fmt and not "svg" in fmt and not svg_already_exists:
-            Path(f"{filename}.svg").unlink()
+        if "html" in fmt and not "svg" in fmt:
+            # SVG file was just needed to generate HTML
+            Path(f"{filename}.tmp.svg").unlink()
+        elif "svg" in fmt:
+            Path(f"{filename}.tmp.svg").replace(f"{filename}.svg")
 
     def bom(self):
         if not self._bom:
--- a/src/wireviz/svgembed.py
+++ b/src/wireviz/svgembed.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+import base64
+import re
+from pathlib import Path
+from typing import Union
+
+mime_subtype_replacements = {"jpg": "jpeg", "tif": "tiff"}
+
+
+def embed_svg_images(svg_in: str, base_path: Union[str, Path] = Path.cwd()) -> str:
+    images_b64 = {}  # cache of base64-encoded images
+
+    def image_tag(pre: str, url: str, post: str) -> str:
+        return f'<image{pre} xlink:href="{url}"{post}>'
+
+    def replace(match: re.Match) -> str:
+        imgurl = match["URL"]
+        if not imgurl in images_b64:  # only encode/cache every unique URL once
+            imgurl_abs = (Path(base_path) / imgurl).resolve()
+            image = imgurl_abs.read_bytes()
+            images_b64[imgurl] = base64.b64encode(image).decode("utf-8")
+        return image_tag(
+            match["PRE"] or "",
+            f"data:image/{get_mime_subtype(imgurl)};base64, {images_b64[imgurl]}",
+            match["POST"] or "",
+        )
+
+    pattern = re.compile(
+        image_tag(r"(?P<PRE> [^>]*?)?", r'(?P<URL>[^"]*?)', r"(?P<POST> [^>]*?)?"),
+        re.IGNORECASE,
+    )
+    return pattern.sub(replace, svg_in)
+
+
+def get_mime_subtype(filename: Union[str, Path]) -> str:
+    mime_subtype = Path(filename).suffix.lstrip(".").lower()
+    if mime_subtype in mime_subtype_replacements:
+        mime_subtype = mime_subtype_replacements[mime_subtype]
+    return mime_subtype
+
+
+def embed_svg_images_file(
+    filename_in: Union[str, Path], overwrite: bool = True
+) -> None:
+    filename_in = Path(filename_in).resolve()
+    filename_out = filename_in.with_suffix(".b64.svg")
+    filename_out.write_text(
+        embed_svg_images(filename_in.read_text(), filename_in.parent)
+    )
+    if overwrite:
+        filename_out.replace(filename_in)
--- a/src/wireviz/wv_html.py
+++ b/src/wireviz/wv_html.py
@@ -37,7 +37,7 @@ def generate_html_output(
     html = open_file_read(templatefile).read()
 
     # embed SVG diagram
-    with open_file_read(f"{filename}.svg") as file:
+    with open_file_read(f"{filename}.tmp.svg") as file:
         svgdata = re.sub(
             "^<[?]xml [^?>]*[?]>[^<]*<!DOCTYPE [^>]*>",
             "<!-- XML and DOCTYPE declarations from SVG file removed -->",


formatc1702:
It seems GraphViz only links to the image in the SVG output. This might be problematic if the output is stored in a different directory from the input, or moved afterwards. Further investigation is required.

formatc1702:
@kvid I'd be interesting in your input on this topic. The solution would be complete if it wasn't for the fact that the SVG, and by extension, the HTML output, depend on having a working link to the original image file. The `.gv` file is also affected, but this should be less of a problem for the average user, since this is just an intermediate file.

formatc1702:
- HTML `` tags allow embedding base64-encoded images within the `SRC` attribute ([example](http://jsfiddle.net/hpP45/)), avoiding the need to link to external files.
- GraphViz does *not* seem to support this. A quick test reveals that
  ```
  attribs['image']['src'] = 'data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=='
  ```
  produces an error:
  ```
  graphviz.backend.CalledProcessError: Command '['dot', '-Tpng', '-O', 'ex08']' returned non-zero exit status 1. [stderr: b'Warning: No such file or directory while opening data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==\nError: No or improper image file="data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=="\nin label of node Key\nWarning: No such file or directory while opening data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==\nError: No or improper image file="data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=="\nin label of node W1\n']
  ```
- Inserting the data manually into an SVG file's `<image xlink:href="">` attribute works! 🎉 
  [Here](https://gist.github.com/formatc1702/afeb1ed01e696cfbfb60bde25f65a019) is `ex08.svg` with both `xlink:href` attributes containing the original images in base64 encoding. Note that the image renders correctly, even though the gist does not include any bitmap files.

@kvid do you think it's a reasonable approach to convert and inject the images into the output SVG file in this way? If so, this PR should only fix the relative path issue, and I'll begin another one to work on the encoding.

kvid:
I'm a bit busy these days, and are not able to contribute as often as I want.

> Here are some new discoveries that might fix the above mentioned problem with embedded images.
> 
> * HTML `` tags allow embedding base64-encoded images within the `SRC` attribute, avoiding the need to link to external files: [example 1](http://jsfiddle.net/hpP45/); [example 2](https://jsfiddle.net/m8ycrn5o/) (encoded by me using [base64-image.de](https://www.base64-image.de/))

That looks useful, but I don't think it's much in use. I've never seen it in an HTML page.

> * GraphViz [does _not_ seem to support this](https://stackoverflow.com/questions/8518070/data-uri-and-graphviz). A quick test (using example 1 above) reveals that
>   ```
>   attribs['image']['src'] = 'data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=='
>   ```
>   
>   
>   produces an error:
>   ```
>   graphviz.backend.CalledProcessError: Command '['dot', '-Tpng', '-O', 'ex08']' returned non-zero exit status 1. [stderr: b'Warning: No such file or directory while opening data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==\nError: No or improper image file="data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=="\nin label of node Key\nWarning: No such file or directory while opening data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==\nError: No or improper image file="data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=="\nin label of node W1\n']
>   ```
> * Inserting the data manually into an SVG file's `<image xlink:href="">` attribute works! 🎉
>   [Here](https://gist.github.com/formatc1702/afeb1ed01e696cfbfb60bde25f65a019) is `ex08.svg` with both `xlink:href` attributes containing the original images in base64 encoding. Note that the image renders correctly, even though the gist does not include any bitmap files.
> * This would fix the problem for SVG, and by extension, HTML output. GraphViz output can't be fixed, but I see this as a minor problem.
> * [More about data URIs](https://en.wikipedia.org/wiki/Data_URI_scheme)
> 
> @kvid do you think it's a reasonable approach to convert and inject the images into the output SVG file in this way? If so, this PR should only fix the relative path issue, and I'll begin another one to work on the encoding.

formatc1702:
> I'm a bit busy these days, and are not able to contribute as often as I want.

Don't worry about it, I know the situation.

You don't seem to have an e-mail address associated with your GitHub account (besides `kvid@users.noreply.github.com`, which I assume won't work)... is there a way for me to send you a direct message? Please let me know via `github@danielrojas.net`.. thanks!

kvid:
The different approaches each have their advantages and disadvantages:
- When specifying an output location different from the input, all image files need to have absolute paths or optionally a modified relative path (if possible) or optionally a copy of the image files can be made at the output location. Converting to absolute paths will not work if the output location is not the final destination. I guess the user should have all 3 options.
- Some users might find it useful to allow a data URI as image source input, but it must be converted to a file for Graphviz to find it, and as in the previous point, absolute path or relative path must be selected.
- Some users might find it useful to generate a data URI in any image source of the SVG output, but I guess it should be optional per image independently of the options above (and maybe a harness option to be used when not specified for each image). The clean way is to make this a new output (e.g. `.data.svg`) to avoid renaming or caching the Graphviz generated `.svg` before overwriting the same output name with the modified version containing data URIs.

formatc1702:
Ideally, the output files should be self-contained, independent of any paths to other files. Anything else just causes more trouble, and the added effort 

Therefore, my final goal is to:
- allow user to specify paths to images inside the YAML, relative (to the YAML file) or absolute
  - embedding data URIs from the start was the idea behind #188 but I don't see it as a priority
- resolve absolute image paths when parsing the YAML
- generate BOM, PNG and SVG output files
- modify SVG output, embedding the images in the `<image xlink:href="">` tags as described above using base64 encoding.
- generate HTML output, containing the modified SVG

formatc1702:
Embedding images in the output SVG works :tada:

It has been [tested](https://gist.github.com/formatc1702/afeb1ed01e696cfbfb60bde25f65a019#file-ex08-svg) with PNG, JPEG and animated GIFs.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,39 @@
...
+    filename_in = Path(filename_in).resolve()
+    filename_out = Path(f'{filename_in.with_suffix("")}.b64.svg')
+    images_b64 = {}  # cache base-64 encoded images
+    re_xlink=re.compile(r"xlink:href=\"(?P<URL>.*?)\"", re.IGNORECASE)
+    with open(filename_in,'r') as file_in, open(filename_out,'w') as file_out:
+        for line in file_in:
+            for xlink in re_xlink.finditer(line):
+                imgurl = xlink.group('URL')
+                if not imgurl in images_b64:
+                    if not Path(imgurl).is_absolute():  # resolve relative image path
```
formatc1702, line 20: Even though resolves relatives paths relative to the input YAML file during parsing (`wireviz.py:L49) I decided to keep this feature here, to make the `embed_svg_images()` function more independent from the WireViz implementation. It would therefore resolve any other remaining paths relative to the SVG file.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,39 @@
...
+
+                    with open(imgurl_abs,'rb') as img:
+                        data_bin = img.read()
+                        data_b64 = base64.b64encode(data_bin)
+                        data_str = data_b64.decode('utf-8')
+                    images_b64[imgurl] = data_str
+
+                mime_subtype = str(imgurl_abs.suffix[1:]).lower()  # remove `.`
+                if mime_subtype in mime_subtype_replacements:
+                    mime_subtype = mime_subtype_replacements[mime_subtype]
```
formatc1702, line 33: This could have been simplified to
```python
mime_subtype = mime_subtype_replacements.get(mime_subtype, mime_subtype)
```
but I decided against it for readability.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,48 @@
...
+    # first, find any image references in SVG data, and cache the respective base64-encoded image
+    images_b64 = {}  # cache of base64-encoded images
+    re_xlink=re.compile(r"xlink:href=\"(?P<URL>.*?)\"", re.IGNORECASE)
+    for xlink in re_xlink.finditer(svg_in):
+        imgurl = xlink.group('URL')
+        if not imgurl in images_b64:  # only encode/cache every unique URL once
+            if not Path(imgurl).is_absolute():
+                imgurl_abs = (Path(base_path) / imgurl).resolve()
+            else:
+                imgurl_abs = Path(imgurl)
```
formatc1702, line 20: Even though WireViz resolves relative paths as relative to the input YAML file during parsing (`wireviz.py:L49`), I decided to keep this feature here, to make the `embed_svg_images()` function more independent from the WireViz implementation. It would therefore resolve any other remaining paths as relative to a specified path, if called from another context.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,48 @@
...
+    if mime_subtype in mime_subtype_replacements:
+        mime_subtype = mime_subtype_replacements[mime_subtype]
+    return mime_subtype
+
+
+def embed_svg_images_file(filename_in: Path, overwrite: bool = True):
+    filename_in = Path(filename_in).resolve()
+    filename_out = Path(f'{filename_in.with_suffix("")}.b64.svg')
+    with open(filename_in,'r') as file_in, open(filename_out,'w') as file_out:
+        file_out.write(embed_svg_images(file_in.read(), filename_in.parent))
```
formatc1702, line 46: When reading from an existing file, interpret relative image paths as relative to the input file.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,48 @@
...
+    for url, b64 in images_b64.items():
+        svg_out = svg_out.replace(url, f'data:image/{get_mime_subtype(url)};base64, {b64}')
+
+    return svg_out
+
+
+def get_mime_subtype(filename: Path):
+    mime_subtype = str(Path(filename).suffix[1:]).lower()  # remove `.`
+    if mime_subtype in mime_subtype_replacements:
+        mime_subtype = mime_subtype_replacements[mime_subtype]
```
formatc1702, line 38: This could have been simplified to
```python
mime_subtype = mime_subtype_replacements.get(mime_subtype, mime_subtype)
```
but I decided against it for readability.

formatc1702:
@kvid, this PR is ready for review.
I am open to discussing your ideas regarding finer user control over how to handle images (embed or not, copy to output directory, ...) but I would first like to have embedded images as default since leaves no open questions.
Thanks!

kvid (changes requested):
I'm sorry this review took so long time. I only had time for a minor section each day, and some of the days, I had to give up due to extremely slow response from github servers.

> @kvid, this PR is ready for review.
> Feel free to mark my comments as resolved if you agree, and let me know if you don't.

I'm not allowed to mark your code comments as resolved, but I've tagged with 👍 where I agree.

> I am open to discussing your ideas regarding finer user control over how to handle images (embed or not, copy to output directory, ...) as a separate PR, but I would first like to have embedded images as default since it leaves no open questions (see my [comment above](https://github.com/formatc1702/WireViz/pull/189#issuecomment-727274871)).
> Thanks!

I agree those other ideas can be handled later. Probably will most users will be happy with this PR and might not give my other ideas a high priority.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,48 @@
...
+
+            with open(imgurl_abs,'rb') as img:
+                img_bin = img.read()
+                img_b64 = base64.b64encode(img_bin)
+                img_str = img_b64.decode('utf-8')
+            images_b64[imgurl] = img_str
+    # second, replace links with the base64-encoded data
+    svg_out = svg_in
+    for url, b64 in images_b64.items():
+        svg_out = svg_out.replace(url, f'data:image/{get_mime_subtype(url)};base64, {b64}')
```
kvid, line 30: A better alternative to the sorting above, is to include known characters before and after the URL:
```suggestion
        svg_out = svg_out.replace(f'href="{url}"', f'href="data:image/{get_mime_subtype(url)};base64, {b64}"')
```

kvid (changes requested):
I have included a few more suggestions, including a fix for a bug that might lead to bad replacements if one URL is a substring of another, and the former is replaced before the latter. The error is that only the substring of the latter will be replaced.

src/wireviz/svgembed.py
formatc1702: See my [comment](https://github.com/formatc1702/WireViz/pull/189#discussion_r524503407) above.

> to make the `embed_svg_images()` function more independent from the WireViz implementation. It would therefore resolve any other remaining paths as relative to a specified path, if called from another context.

This line will never get called if coming from the WireViz parser, but I can see the SVG image embedding being useful in other contexts (maybe as a separate package/module), so I kept it for completeness sake.

src/wireviz/svgembed.py
formatc1702: Thanks. I'll leave it as it is now :)

src/wireviz/Harness.py
```diff
@@ -342,12 +343,8 @@ def png(self):
 
     @property
     def svg(self):
-        from io import BytesIO
         graph = self.create_graph()
-        data = BytesIO()
-        data.write(graph.pipe(format='svg'))
-        data.seek(0)
-        return data.read()
+        return embed_svg_images(graph.pipe(format='svg').decode('utf-8'), Path.cwd())
```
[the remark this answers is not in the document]
formatc1702: I figured, if the `svg()` property is called when using WireViz as a module, and creating a `Harness` object directly in code, as opposed to from an input YAML file, `cwd` would be the most reasonable assumption for resolving relative image paths.

I haven't used this feature myself, so perhaps there is a better default value for the `base_path` agrument? I am open for suggestions.

src/wireviz/svgembed.py
formatc1702: How about using only `f'="{url}"'`, without `href`, to avoid any upper/lower case issues?
The sorting by length using lambda is clever, but a bit cryptic; this approach seems more straightforward :)

src/wireviz/wireviz.py
```diff
@@ -17,7 +17,7 @@
 from wireviz.wv_helper import expand, open_file_read
 
 
-def parse(yaml_input: str, file_out: (str, Path) = None, return_types: (None, str, Tuple[str]) = None) -> Any:
+def parse(yaml_input: str, file_in: (str, Path) = None, file_out: (str, Path) = None, return_types: (None, str, Tuple[str]) = None) -> Any:
```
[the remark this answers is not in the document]
formatc1702: Any reason why `Union[]` is preferred over `()` for type hints?
Especially since, AFAIK, they are not even enforced by Python, but are more of a documentation feature?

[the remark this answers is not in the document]
formatc1702: Also, what's the meaning of `Tuple[str, ...]` towards the end of your suggestion?

src/wireviz/wireviz.py
```diff
@@ -17,7 +17,7 @@
...
     """
     Parses yaml input string and does the high-level harness conversion
```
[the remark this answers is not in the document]
formatc1702: I've changed `file_in` to `base_path` in 6a42a30523f49fa894d7c4e694319f355eddc972, please see the commit message for the reasoning. The param list has been updated accordingly.

formatc1702:
Thanks for your suggestions. Don't worry about not responding earlier.. there is no obligation here.
I've included most of your suggestions and commented on the ones that need discussion.

src/wireviz/wireviz.py
```diff
@@ -209,7 +210,7 @@ def parse_file(yaml_file: str, file_out: (str, Path) = None) -> None:
         file_out = fn
     file_out = os.path.abspath(file_out)
 
-    parse(yaml_input, file_out=file_out)
+    parse(yaml_input, file_in=Path(yaml_file).resolve(), file_out=file_out)
```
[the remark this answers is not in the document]
formatc1702: See 6a42a30523f49fa894d7c4e694319f355eddc972 for the new implementation and let me know what you think!

src/wireviz/wireviz.py
```diff
@@ -251,7 +252,7 @@ def main():
         file_out = args.output_file
     file_out = os.path.abspath(file_out)
 
-    parse(yaml_input, file_out=file_out)
+    parse(yaml_input, file_in=Path(args.input_file).resolve(), file_out=file_out)
```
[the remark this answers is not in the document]
formatc1702: See 6a42a30523f49fa894d7c4e694319f355eddc972 for the new implementation and let me know what you think!

src/wireviz/wireviz.py
```diff
@@ -44,10 +44,11 @@ def parse(yaml_input: str, file_out: (str, Path) = None, return_types: (None, st
...
                     for key, attribs in yaml_data[sec].items():
-                        # The Image dataclass might need to open an image file with a relative path.
-                        image = attribs.get('image')
-                        if isinstance(image, dict):
-                            image['gv_dir'] = Path(file_out if file_out else '').parent # Inject context
+                        if attribs.get('image'):
+                            image_path = attribs['image']['src']
+                            if not Path(image_path).is_absolute():  # resolve relative image path
+                                image_path = (Path(file_in).parent / image_path).resolve()
+                                attribs['image']['src'] = image_path
```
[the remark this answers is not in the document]
formatc1702: See 6a42a30523f49fa894d7c4e694319f355eddc972, maybe this gets closer to what you propose.

src/wireviz/wireviz.py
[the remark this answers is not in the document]
kvid: I know type hints are not enforced, but if we are going to use it in this project, I vote for using the same syntax where possible, and I have not seen any documentation of  `()` for type hints with alternative types, only `Union[]` - se links below:
- Background with PEP-links and video presentation by Guido van Rossum: [What are type hints in Python 3.5?](https://stackoverflow.com/questions/32557920/what-are-type-hints-in-python-3-5)
- Doc: [typing library doc](https://docs.python.org/3/library/typing.html)
- Quick overview: [Type hints cheat sheet (Python 3)](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html)
```py
# For tuples of fixed size, we specify the types of all the elements
x: Tuple[int, str, float] = (3, "yes", 7.5)

# For tuples of variable size, we use one type and ellipsis
x: Tuple[int, ...] = (1, 2, 3)
```

src/wireviz/Harness.py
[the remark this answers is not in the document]
kvid: OK. No alternative suggestions for the moment. I have not yet seen any code using WireViz as a module.

src/wireviz/wireviz.py
```diff
@@ -44,10 +45,11 @@ def parse(yaml_input: str, file_out: (str, Path) = None, return_types: (None, st
...
                     for key, attribs in yaml_data[sec].items():
-                        # The Image dataclass might need to open an image file with a relative path.
-                        image = attribs.get('image')
-                        if isinstance(image, dict):
-                            image['gv_dir'] = Path(file_out if file_out else '').parent # Inject context
+                        if attribs.get('image'):
+                            image_path = attribs['image']['src']
+                            if not Path(image_path).is_absolute():  # resolve relative image path
+                                image_path = (Path(base_path) / image_path).resolve()
+                                attribs['image']['src'] = image_path
```
kvid, line 52: I've already suggested an alternative to this code in https://github.com/formatc1702/WireViz/pull/189#discussion_r532944282. Why do you seem to prefer repeating lookup the `'image'` key 3 times instead of assigning and reusing an image variable?

kvid (changes requested):
> Thanks for your suggestions. Don't worry about not responding earlier.. there is no obligation here.
> I've included most of your suggestions and commented on the ones that need discussion/clarification.

Thank you for accepting most of my suggestions. I've tried answering the issues with questions, and added some more suggestions as well.

src/wireviz/Harness.py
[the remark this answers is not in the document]
formatc1702: I'd rather require explicitly supplying this argument to "force" thinking about the issue when using this function elsewhere.

src/wireviz/svgembed.py
formatc1702: That is a valid point.
How about a Regex that picks out `href="*"` only within `<a *>`, to avoid the need for a parser as additional dependency?
Sure, a parser may work better for some ultra-rare edge case, but it's not worth it IMHO.

formatc1702: How about a case-independent Regex that matches `xlink:href="*"`?
It would ignore a regular `<a href="*">` and wouldn't bloat the project with a parser as an additional dependency.
Sure, there might be ultra-rare cases that the Regex couldn't handle as well as a parser, but I'm willing to take that risk ;-)
What do you think?

src/wireviz/wireviz.py
```diff
@@ -44,10 +45,11 @@ def parse(yaml_input: str, file_out: (str, Path) = None, return_types: (None, st
             if len(yaml_data[sec]) > 0:
                 if ty == dict:
                     for key, attribs in yaml_data[sec].items():
-                        # The Image dataclass might need to open an image file with a relative path.
                         image = attribs.get('image')
                         if isinstance(image, dict):
-                            image['gv_dir'] = Path(file_out if file_out else '').parent # Inject context
+                            image_path = image['src']
+                            if image_path and not Path(image_path).is_absolute():  # resolve relative image path
+                                image['src'] = (Path(base_path) / image_path).resolve()
```
formatc1702, line 52: Thanks @kvid, I've included the suggestion from your [comment](https://github.com/formatc1702/WireViz/pull/189#discussion_r532944282).

I'm a little confused why `image` is a *reference* to `attribs` and therefore assigning `image[src] = ...` directly affects `yaml_data`'s contents; where as `image_path` just gets passed the *value* of the dict key, and `image_path = ...` does not affect `yaml_data`.

Even if using `image_path = image.get('src')` in line 50 (analogously to `image = attribs.get('image')` in line 48), you cannot do the final assignment with `image_path = (Path(base_path) / image_path).resolve()`.

Maybe I'm missing something... can you explain why `image` is a reference and `image_path` is not? Thanks :)

kvid: > Thanks @kvid, I've included the suggestion from your [comment](https://github.com/formatc1702/WireViz/pull/189#discussion_r532944282).

Thank you for accepting my suggestion.

> 
> I'm a little confused why `image` is a _reference_ to `attribs` and therefore assigning `image[src] = ...` directly affects `yaml_data`'s contents; where as `image_path` just gets passed the _value_ of the dict key, and `image_path = ...` does not affect `yaml_data`.
> 
> Even if using `image_path = image.get('src')` in line 50 (analogously to `image = attribs.get('image')` in line 48), you cannot do the final assignment with `image_path = (Path(base_path) / image_path).resolve()`.
> 
> Maybe I'm missing something... can you explain why `image` is a reference and `image_path` is not? Thanks :)

Both `image` and `image_path` are references to objects that also are reachable through `yaml_data`.
- The `image` variable is a reference to a `dict` that is mutable, and this `dict` can be changed through the reference, e.g. `image['src'] = 'new value'`, and the changed object is also are reachable through `yaml_data` (the reference is unchanged).
- However, The `image_path` variable is a reference to a `str` that is immutable, and immutable objects cannot be changed. Assigning a different value to the `image_path` variable will make it a reference to a different object, but that does not affect the original referenced immutable object (that is still reachable through `yaml_data`).

I hope my description above is somewhat understandable, and don't create even more confusion. Others have described this differences in more detail:
- https://www.freecodecamp.org/news/mutable-vs-immutable-objects-python/
- https://medium.com/@meghamohan/mutable-and-immutable-side-of-python-c2145cf72747

formatc1702: Thanks for the explanation and links! Very helpful. Will mark as resolved.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,40 @@
...
+        imgurl = xlink.group('URL')
+        if not imgurl in images_b64:  # only encode/cache every unique URL once
+            imgurl_abs = (Path(base_path) / imgurl).resolve()
+            images_b64[imgurl] = base64.b64encode(imgurl_abs.read_bytes()).decode('utf-8')
+    # second, replace links with the base64-encoded data
+    svg_out = svg_in
+    for url, b64 in images_b64.items():
+        svg_out = svg_out.replace(url, f'data:image/{get_mime_subtype(url)};base64, {b64}')
+
+    return svg_out
```
kvid, line 25: To reduce the probability of false replacements, I suggest searching for the full `<image>` tag including any attributes preceding and/or succeeding `xlink:href`, and use `sub()` to replace all such URLs in one pass instead of separate loops for find and replace:
```suggestion
    images_b64 = {}  # cache of base64-encoded images

    def image_tag(pre: str, url: str, post: str) -> str:
        return f'<image{pre} xlink:href="{url}"{post}>'

    def replace(match: re.Match) -> str:
        imgurl = match['URL']
        if not imgurl in images_b64:  # only encode/cache every unique URL once
            imgurl_abs = (Path(base_path) / imgurl).resolve()
            images_b64[imgurl] = base64.b64encode(imgurl_abs.read_bytes()).decode('utf-8')
        return image_tag(
            match['PRE'] or '',
            f'data:image/{get_mime_subtype(imgurl)};base64, {images_b64[imgurl]}',
            match['POST'] or '')

    pattern = re.compile(
        image_tag(r'(?P<PRE> [^>]*?)?', r'(?P<URL>[^"]*?)', r'(?P<POST> [^>]*?)?'), re.IGNORECASE)
    return pattern.sub(replace, svg_in)
```

src/wireviz/svgembed.py
kvid: > How about a case-independent Regex that matches `xlink:href="*"`?
> It would ignore a regular `<a href="*">` and wouldn't bloat the project with a parser as an additional dependency.

They will look like `<a xlink:href="*">` and there might also be additional attributes in the same tag preceding or succeeding `xlink:href`. To find only `<image>` tags using regular expressions, I suggest searching for the full tag.

> Sure, there might be ultra-rare cases that the Regex couldn't handle as well as a parser, but I'm willing to take that risk ;-)
> What do you think?

See my rewrite suggestion to reduce the risk.

kvid (changes requested):
A major drawback with resolving all image paths to the absolute path, is that the output file contents will depend on the absolute path that normally differ between developers, and thereby lead to false diffs when doing regression testing.

See also suggested improvement of `embed_svg_images()`.

formatc1702:
> A major drawback with resolving all image paths to the absolute path, is that the output file contents will depend on the absolute path that normally differ between developers, and thereby lead to false diffs when doing regression testing.

The way I see it, this would only affect `.gv` files.

- Is it worth thinking about a solution in code to output a version of the `.gv` file with relative paths (for the diff), and one with absolutes?
- Is it possible to somehow ignore certain patterns (i.e. only the `src` attribute of an `` tag) while doing the diff?
  - Personally, I would prefer this option, and it could be a task to include in #63 ...

src/wireviz/svgembed.py
formatc1702: This looks promising.
I'll need a bit of time to wrap my head around the logic; passing three `r''` strings to a function, and then passing the entire `replace` function to `pattern.sub()`...
It's a bit hard to understand; is that the most pythonic way to do it? I'm not complaining.. the overall approach of filtering the `<image>` tag and only replacing `xlink:href` within it seems like a very good solution.

src/wireviz/svgembed.py
```diff
@@ -0,0 +1,48 @@
...
+                imgurl_abs = Path(imgurl)
+
+            with open(imgurl_abs,'rb') as img:
+                img_bin = img.read()
+                img_b64 = base64.b64encode(img_bin)
+                img_str = img_b64.decode('utf-8')
+            images_b64[imgurl] = img_str
+    # second, replace links with the base64-encoded data
+    svg_out = svg_in
+    for url, b64 in images_b64.items():
```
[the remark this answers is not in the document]
formatc1702: Marking as resolved, following the regex approach below instead.

src/wireviz/svgembed.py
formatc1702: Marking as resolved, following the regex approach below instead.

src/wireviz/svgembed.py
kvid: > This looks promising.
> I'll need a bit of time to wrap my head around the logic; passing three `r''` strings to a function,

If you see an alternative way to avoid repeating the `<image... xlink:href="..."...>` tag string when needing it both in the search pattern and in the replacement string, please suggest something that might be better. Maybe renaming some identifiers might help the readability? I agree rhe `pre` and `post` names are not very good.

> and then passing the entire `replace` function to `pattern.sub()`...

Using a replacement function instead of a replacement string is the thing to do when needing some processing of the matching string to create the replacement.

> It's a bit hard to understand; is that the most pythonic way to do it?

What is easy to understand might depend on what you are used to see. What is most Pythonic might also differ a bit between developers, I guess. I know there are PEPs describing what to prefer in many situations, but they cannot cover everything.

> I'm not complaining.. the overall approach of filtering the `<image>` tag and only replacing `xlink:href` within it seems like a very good solution.

Thank's

kvid:
> > A major drawback with resolving all image paths to the absolute path, is that the output file contents will depend on the absolute path that normally differ between developers, and thereby lead to false diffs when doing regression testing.
> 
> The way I see it, this would only affect `.gv` files.

I agree.

> 
> * Is it worth thinking about a solution in code to output a version of the `.gv` file with relative paths (for the diff), and one with absolutes?

This is probably the easiest to implement.

> * Is it possible to somehow ignore certain patterns (i.e. only the `src` attribute of an `` tag) while doing the diff?
>   
>   * This could be a task to include in #63 ...

This would probably need to involve temporary renaming, create modified version, doing diff, and rename back for each `.gv` file, and something similar for a custom `git add` function to create the comparing base. It seems more complicated than your first suggestion.

> 
> Once #60 is finally implemented, `.gv` will likely not be used by most people outside of WireViz development.

I don't think it is used much by most people outside of WireViz development today either...

formatc1702:
What would be the best way of implementing separate `.gv` files?
Some thoughts:

- Image paths are resolved in `wireviz.py -> parse()`, since this is the context where the base path is known.
  - I propose reading the `image.src` attribute here, and placing the absolute path in a new `image.src_abs` attribute.
- The content for the`.gv` file is generated in `Harness.py -> create_graph()`, which uses `wv_gv_html.py -> html_image()`
  - How about a flag for `create_graph()` to choose whether to use `image.src` or `image.src_abs` when calling `html_image()`?
- The actual `.gv` file is output via the `Harness.py -> output()` function.
  - It feels kind of wasteful to call `create_graph()` twice just to get the two file variants.
  - One option is to use the `fmt` parameter to decide which file to generate. `gv_rel` and `gv_abs` could be two of the allowed options and trigger the respective output. This way, while developing, both files can be generated, while no double effort is wasted during normal use.

What do you think?

kvid:
I'm sorry about this late answer. I agree with the suggested way of implementing separate `.gv` files. It seems like a reasonable approach.

A similar approach could also be used to enable an optional `.html` output with an ordinary image tag instead of embedding the `.svg` file. Then it will be possible to do regression testing on that file too.

src/wireviz/wireviz.py
```diff
@@ -251,7 +253,7 @@ def main():
         file_out = args.output_file
     file_out = os.path.abspath(file_out)
 
-    parse(yaml_input, file_out=file_out)
+    parse(yaml_input, base_path=Path(args.input_file).parent, file_out=file_out)
```
kvid, line 256: This `yaml_input` might be a concatenation of a prepended file and the main input file, and these two files might not be located in the same folder. Any image with relative path in the prepended YAML file might in that case not be found using the `base_path` from the main input file only. See use case in https://github.com/formatc1702/WireViz/issues/220#issuecomment-797154027 by @gmkado.

I suggest solving this in a way that might also support a set of prepended files.

formatc1702: How about resolving paths in the prepended file and the main file separately, before concatenating? I am open for a new PR, based on this one, showing the implementation. Again, apologies for not tackling this myself.

kvid: - I have not found a way to parse the files separately (with different base paths) and then join them into the same document, except maybe (load+resolve+dump) each file and then join the results before loading the combined dumps, but I have not tested something like this, and I don't know if some information might get lost in the process.
- A completely different approach is to put the different base paths into a path list used as the value of the [imagepath](https://graphviz.org/doc/info/attrs.html#d:imagepath) attribute, keep each image path relative, and trust Graphviz to resolve the paths. The path list should be in reverse order so the base path of the main file is first, because if the same image file name is available from more than one base path, the one in the base path of the main file should have priority. A side-effect is that this approach might remove the need for two separate `.gv` files (or at least make it easier to create the second from the first without generating everything twice).
- The same path list can of course be used to resolve each image path in WireViz, but then we loose the side-effect.

formatc1702:
> > > A major drawback with resolving all image paths to the absolute path, is that the output file contents will depend on the absolute path that normally differ between developers, and thereby lead to false diffs when doing regression testing.
> > 
> > 
> > The way I see it, this would only affect `.gv` files.
> 
> I agree.

I am working on a refactoring of the WireViz CLI in #244 which will finally close #60 and will stop outputting `.gv` files by default.
Since you agree that `.gv` is rarely needed outside development, would you be OK if I implement your other suggestions, ignore the fact that absolute paths will be visible if `.gv` output is enabled, and we get this PR ready?
I am sure we can then create a follow-up PR to elegantly deal with the case where `.gv` *is* generated.

src/wireviz/wireviz.py
formatc1702: See my [comment in the main discussion](https://github.com/formatc1702/WireViz/pull/189#issuecomment-932752516) and please mark resolved if you agree.

formatc1702: The GraphViz `imagepath` approach sounds nice in theory, but when you start needing a different separator depending on the OS, it gets ugly fast.

formatc1702:
Rebased on top of [`latest`](https://github.com/formatc1702/WireViz/tree/latest) to profit from `smart_file_resolve()`.
Old branch with full commit history is backed up as [`backup/bugfix/image-src_before-latest-rebase`](https://github.com/formatc1702/WireViz/tree/backup/bugfix/image-src_before-latest-rebase)

PR #192, opened 2020-11-01
Move BOM management and HTML functions to separate modules

### LINKED ISSUE #151: [internal] Create a new module for helper functions to each output format?

formatc1702:
Closes #151.

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -4,6 +4,7 @@
 from typing import Optional, List, Tuple, Union
 from dataclasses import dataclass, field, InitVar
 from pathlib import Path
+
 from wireviz.wv_helper import int2tuple, aspect_ratio
 from wireviz import wv_colors
 
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -1,19 +1,23 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-from wireviz.DataClasses import Connector, Cable
 from graphviz import Graph
-from wireviz import wv_colors, wv_helper, __version__, APP_NAME, APP_URL
-from wireviz.wv_colors import get_color_hex
-from wireviz.wv_helper import awg_equiv, mm2_equiv, tuplelist2tsv, \
-    nested_html_table, flatten2d, index_if_list, html_line_breaks, \
-    clean_whitespace, open_file_read, open_file_write, html_colorbar, \
-    html_image, html_caption, manufacturer_info_field, component_table_entry, remove_links
 from collections import Counter
 from typing import List, Union
 from pathlib import Path
 from itertools import zip_longest
 import re
+
+from wireviz import wv_colors, __version__, APP_NAME, APP_URL
+from wireviz.DataClasses import Connector, Cable
+from wireviz.wv_colors import get_color_hex
+from wireviz.wv_gv_html import nested_html_table, html_colorbar, html_image, \
+    html_caption, remove_links, html_line_breaks
+from wireviz.wv_bom import manufacturer_info_field, component_table_entry, \
+    get_additional_component_table, bom_list, generate_bom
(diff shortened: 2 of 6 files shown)


formatc1702:
Moving the BOM modules to a separate file might be trickier than expected...

kvid:
I know this is WIP, but in case these hints might help you:
- `get_additional_component_table` and `manufacturer_info_field` is not used for BOM and might rather belong to `wv_gv_html.py`.
- `clean_whitespace` is not used for diagram output. It is rather generic, but currently used for BOM.
- `remove_links` is rather generic and mainly used for diagram output, but also for `.bom.tsv` output.

formatc1702:
Thanks, yes, I'm moving things around as I progress.

Currently, `build_examples.py` runs without errors, but BOMs are empty.

It seems things are getting lost inside `generate_bom()` when passing into from the `bom_entries` variable to `bom`.

I think the BOM generation code has grown quite complex, so I need some time to understand it, and make sure we are not passing around more objects than needed, that functions end up in the correct modules, etc.

formatc1702:
> * `get_additional_component_table` and `manufacturer_info_field` is not used for BOM and might rather belong to `wv_gv_html.py`.

- `get_additional_component_table` is tightly linked to `get_bom_index()` and that one to the actual `bom()` function, therefore I will choose to leave it in `wv_bom.py`, at least until we refactor the BOM code, which should be a separate PR.
- `manufacturer_info_field` is also used in `component_table_entry` for generating the mini-BOM within a component node, and it's related to the typical BOM fields `pn` etc., so I think it also belongs in `wv_bom.py`. Also, it does not generate any actual HTML.

> * `clean_whitespace` is not used for diagram output. It is rather generic, but currently used for BOM.
> * `remove_links` is rather generic and mainly used for diagram output, but also for `.bom.tsv` output.

I agree they're generic enough to move to `wv_helper.py` since they are not really tied to BOM or HTML stuff.

formatc1702:
`build_examples.py` now runs without errors, and the output is exactly the same as when using the `dev` branch.

Therefore, unless no glaring errors show up, I would like to merge this soon. Refactoring and optimization of the affected functions can happen in a later PR. Thanks!

kvid (approved):
I can verify that the output is unchanged when generating examples. I do have some simplifications of `wv_bom.py` in my local repo that I plan to push to a new PR when this is merged, as you requested.

PR #194, opened 2020-11-11
Reference wires by color or label

### LINKED ISSUE #70: [feature] Add ability to show pin function (from pinout list) next to pin number inside wire/bundle
This makes it easier to understand what a wire is connecting, at a glance. Currently, only connector and pin number is shown (e.g. `X1:4`), seeing the function there would be helpful.

### LINKED ISSUE #169: [feature] better readability for cables with numbered wires
I recently created a cable using YY control cable that has numbered cores rather than different colors, this results in needing to count how far down the core is to work out its number. A potential soloution for these cables or others that don't use colour to diferentiate might be to add a wirelabel option similar to the pinlabels used on connectors. What are peoples opinions on this?

### LINKED ISSUE #193: [Feature] Allow referencing a cable's/bundle's wires by color or label, if unambiguous
Similar to how a connector's pins can be referenced by its ID (as per its entry in `pins`), or by its label (as per its entry in `pinlabels`) in case the label is unambiguous, a cable's/bundle's wire needs the ability to be referenced by its color (if unambiguous) since that is often the most natural way for the user. Additionally, cables should have a `wirelabels` attribute analogous to a connector's `pinlabels`, that can also be used for referencing.

[Example 08](https://github.com/formatc1702/WireViz/tree/master/examples#example-08)'s connection set should therefore work with either of the following:

- Using the wire numbers (currently the only way):
  ```yaml
  connections:
    -
      - Key: [S,R,T]
      - W1: [1,2,3]
  ```

- Using the wire colors:
  ```yaml
  connections:
    -
      - Key: [S,R,T]
      - W1: [WH, BN, GN]
  ```

- Using the new `wirelabels` attribute:
  ```yaml
  ...
  
  cables:
    W1:
      color_code: DIN
      wirecount: 3
      wirelabels: [one, two, three]
      ...
  
  connections:
    -
      - Key: [S,R,T]
      - W1: [one, two, three]
  ```

formatc1702:
Closes #70.
Closes #169.
Closes #193.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -35,30 +35,46 @@ class Harness:
     def add_bom_item(self, item: dict) -> None:
         self.additional_bom_items.append(item)
 
-    def connect(self, from_name: str, from_pin: (int, str), via_name: str, via_pin: (int, str), to_name: str, to_pin: (int, str)) -> None:
-        for (name, pin) in zip([from_name, to_name], [from_pin, to_pin]):  # check from and to connectors
+    def connect(self, from_name: str, from_pin: (int, str), via_name: str, via_wire: (int, str), to_name: str, to_pin: (int, str)) -> None:
+        # check from and to connectors
+        for (name, pin) in zip([from_name, to_name], [from_pin, to_pin]):
             if name is not None and name in self.connectors:
                 connector = self.connectors[name]
+                # check if provided name is ambiguous
                 if pin in connector.pins and pin in connector.pinlabels:
-                    if connector.pins.index(pin) == connector.pinlabels.index(pin):
-                        # TODO: Maybe issue a warning? It's not worthy of an exception if it's unambiguous, but maybe risky?
-                        pass
-                    else:
+                    if connector.pins.index(pin) != connector.pinlabels.index(pin):
                         raise Exception(f'{name}:{pin} is defined both in pinlabels and pins, for different pins.')
+                    # TODO: Maybe issue a warning if present in both lists but referencing the same pin?
                 if pin in connector.pinlabels:
                     if connector.pinlabels.count(pin) > 1:
                         raise Exception(f'{name}:{pin} is defined more than once.')
-                    else:
-                        index = connector.pinlabels.index(pin)
-                        pin = connector.pins[index] # map pin name to pin number
-                        if name == from_name:
-                            from_pin = pin
-                        if name == to_name:
-                            to_pin = pin
+                    index = connector.pinlabels.index(pin)
+                    pin = connector.pins[index] # map pin name to pin number
+                    if name == from_name:
+                        from_pin = pin
+                    if name == to_name:
+                        to_pin = pin
                 if not pin in connector.pins:
                     raise Exception(f'{name}:{pin} not found.')
 
-        self.cables[via_name].connect(from_name, from_pin, via_pin, to_name, to_pin)
+        # check via cable
+        if via_name in self.cables:
+            cable = self.cables[via_name]
+            # check if provided name is ambiguous
+            if via_wire in cable.colors and via_wire in cable.wirelabels:
+                if cable.colors.index(via_wire) != cable.wirelabels.index(via_wire):
+                    raise Exception(f'{via_name}:{via_wire} is defined both in colors and wirelabels, for different wires.')
+                # TODO: Maybe issue a warning if present in both lists but referencing the same wire?
+            if via_wire in cable.colors:
+                if cable.colors.count(via_wire) > 1:
+                    raise Exception(f'{via_name}:{via_wire} is used for more than one wire.')
+                via_wire = cable.colors.index(via_wire) + 1  # list index starts at 0, wire IDs start at 1
+            elif via_wire in cable.wirelabels:
+                if cable.wirelabels.count(via_wire) > 1:
+                    raise Exception(f'{via_name}:{via_wire} is used for more than one wire.')
+                via_wire = cable.wirelabels.index(via_wire) + 1  # list index starts at 0, wire IDs start at 1
+
+        self.cables[via_name].connect(from_name, from_pin, via_wire, to_name, to_pin)
         if from_name in self.connectors:
             self.connectors[from_name].activate_pin(from_pin)
         if to_name in self.connectors:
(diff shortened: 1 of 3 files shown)


formatc1702:
I'm sure the `Harness.connect()` function can be simplified and cleaned up quite a bit. My hope is to include this feature soon (we have been needing this at work and I finally found some time for it) and do some cleanup in the upcoming major refactoring.

formatc1702:
Update: Wire number can now be displayed as suggested in #169. It is controlled by the `Cable.show_wirenumber` parameter, which defaults to `True` for "normal" cables (which usually follow a numbering scheme) and to `False` for bundles (which have no particular order), but can be easily overridden.

kvid:
> Update: Wire number can now be displayed as suggested in #169. It is controlled by the `Cable.show_wirenumbers` parameter, which defaults to `True` for "normal" cables (which usually follow a numbering scheme) and to `False` for bundles (which have no particular order), but can be easily overridden.

I suggest also default True when neither wire colors nor wire labels can identify all wires unambiguously, because then wire numbers are the only unambiguous wire identifier.

> 
> Expand to see examples 2 (cables, numbered wires) and 3 (bundle, no numbered wires), as well as a modified example 8 (numbered wires + labels)

I suggest showing any empty color or label as an empty string to avoid skipping a ':' separator in such cases, because if some wires are shown with fewer separators, then it is unambiguous which of the identifiers are empty.

formatc1702:
Update: Adressed #70 as well.

formatc1702:
> > Update: Wire number can now be displayed as suggested in #169. It is controlled by the `Cable.show_wirenumbers` parameter, which defaults to `True` for "normal" cables (which usually follow a numbering scheme) and to `False` for bundles (which have no particular order), but can be easily overridden.
> 
> I suggest also default True when neither wire colors nor wire labels can identify all wires unambiguously, because then wire numbers are the only unambiguous wire identifier.

This would only be the case for bundles with ambiguous colors/labels... in which case, the concept of having a numbering no longer applies IMHO. In that case, both in the diagram and in real life, following a wire from one end to the other is the only way to make sure, and I don't see a need for complicating the program logic here.

> > Expand to see examples 2 (cables, numbered wires) and 3 (bundle, no numbered wires), as well as a modified example 8 (numbered wires + labels)
> 
> I suggest showing any empty color or label as an empty string to avoid skipping a ':' separator in such cases, because if some wires are shown with fewer separators, then it is unambiguous which of the identifiers are empty.

Can you provide a realistic example of where this would be an issue? If a cable has defined colors, surely there would be no gap in the color list? As for labels, if some wires have none, the final `:<label>` goes missing, but the position of the wire number (if shown) and wire color stay consistent.

src/wireviz/Harness.py
```diff
@@ -185,10 +202,22 @@ def create_graph(self) -> Graph:
...
+                wirehtml.append(f'    <td>')
+
+                wireinfo = []
+                if cable.show_wirenumbers:
+                    wireinfo.append(str(i))
+                colorstr = wv_colors.translate_color(connection_color, self.color_mode)
+                if colorstr:
+                    wireinfo.append(colorstr)
+                if cable.wirelabels:
+                    wireinfo.append(wirelabel if wirelabel else '')
```
kvid, line 217:
```suggestion
                    wireinfo.append(wirelabel if wirelabel is not None else '')
```
Otherwise, you might risk hiding labels that has non-string values that evaluates to False, e.g. `0`, `False`, `0.0`, `0j`, etc.

kvid:
> > I suggest also default True when neither wire colors nor wire labels can identify all wires unambiguously, because then wire numbers are the only unambiguous wire identifier.
> 
> This would only be the case for bundles with ambiguous colors/labels... in which case, the concept of having a numbering no longer applies IMHO. In that case, both in the diagram and in real life, following a wire from one end to the other is the only way to make sure,

Well, the numbering still applies as the rendering order in the diagram and for specifying connections - the only way in this case. And in real life there are cases where the wires have an order, either some kind of numeric labels or e.g. the order they are strapped to the surrounding structure.
ordered unicolored wires

However, the user can always specify `Cable.show_wirenumbers: True` manually in such cases.

> and I don't see a need for complicating the program logic here.

That might be a valid argument. The program logic complexity would be at least twice the similar test for unique connector pins, i.e. something like this:
```py
all(len(ids) < self.wirecount or len(ids) != len(set(ids)) for ids in [self.colors, self.wirelabels])
```

> If it's a (non-bundle) cable with ambiguous colors/labels, it is already defaulting to `True`.

That's true.

> 
> > > Expand to see examples 2 (cables, numbered wires) and 3 (bundle, no numbered wires), as well as a modified example 8 (numbered wires + labels)
> > 
> > 
> > I suggest showing any empty color or label as an empty string to avoid skipping a ':' separator in such cases, because if some wires are shown with fewer separators, then it is unambiguous which of the identifiers are empty.
> 
> Can you provide a realistic example of where this would be an issue? If a cable has defined colors, surely there would be no gap in the color list?

You are probably right. After reading your current code changes, I'm not that worried. 😃 

> As for labels, if some wires have none, the final `:<label>` goes missing, but the position of the wire number (if shown) and wire color stay consistent.

When some labels are shown, and some are empty, I would prefer a trailing ':' for the wires with an empty label to easily see where labels are missing, but I accept that you disagree.

> 
> If a cable has wire labels but no set colors, it's more readable to simply leave out the color section altogether, instead of filling it with `--` (or similar) or having a double `:` (e.g. `1::label`) for no real reason...

I agree with you about the cases where the whole list (`cable.colors` or `cable.wirelabels`) is empty.

formatc1702:
> > > I suggest also default True when neither wire colors nor wire labels can identify all wires unambiguously, [...]
> 
> [...] And in real life there are cases where the wires have an order, either some kind of **numeric labels** [...]

`<troll>` I'm afraid you've gone full circle here 😛 `</troll>` (SCNR)

> or e.g. the order they are strapped to the surrounding structure.

That one is a valid point. Let's leave the logic as-is for now and let the user override in this situation.

> When some labels are shown, and some are empty, I would prefer a trailing ':' for the wires with an empty label to easily see where labels are missing, but I accept that you disagree.

I'll think about it; it's not completely unreasonable, but I'm leaning towards leaving the trailing `:` out.

kvid:
> > > > I suggest also default True when neither wire colors **nor wire labels** can identify all wires unambiguously, [...]
> > 
> > [...] And in real life there are cases where the wires have an order, either some kind of **numeric labels** [...]
> 
> `<troll>` I'm afraid you've gone full circle here 😛 `</troll>` (SCNR)

LOL 🤣 I see your point. 😆 I was thinking of a case where the diagram wire labels (and possible physical written labels) represent the wire function, and that it could exist some kind of numeric marks on the wires as well, but a better way to handle that would be to prefix the diagram wire labels with the numeric marks to obtain unambiguous labels.

> 
> > or e.g. the order they are strapped to the surrounding structure.
> 
> That one is a valid point. Let's leave the logic as-is for now and let the user override in this situation.

I'm perfectly fine with that.

> 
> > When some labels are shown, and some are empty, I would prefer a trailing ':' for the wires with an empty label to easily see where labels are missing, but I accept that you disagree.
> 
> I'll think about it; it's not completely unreasonable, but I'm leaning towards leaving the trailing `:` out.

I can live with either solution. It's not a major issue to me, I was only trying to explain the rationale behind my original suggestion in a better way.

formatc1702:
I've rebased onto the latest `dev` after merging #141; hopefully this one can follow soon after :)

kvid:
> I've rebased onto the latest `dev` after merging #141;

👍 

> hopefully this one can follow soon after :)

Do you wait for an OK from me? As your previous comment included "I'll think about it...", then I thought I would wait until you concluded before reviewing again. Have you made any recent changes (other than rebasing)? I don't have my PC here to inspect diffs.

formatc1702:
Sorry for being vague. I prefer leaving the `:` out. I haven't changed anything besides the rebase.
So please have a look of you find any major problems... like I said, once we go object-oriented, a lot of the code will hopefully be simplified, so I wouldn't look to optimize it to perfection this time ;-) Thanks!

kvid (approved):
> Sorry for being vague.

No worries. I'm sorry I misunderstood you. As you requested a quick feedback without perfection in mind, I have mainly rebuilt all examples and the diff seems as expected. Thank you also for accepting my earlier suggestion. I don't have the time to look for simplifications of the connection code right now, but I guess that can be done later. Sorry for not being as thorough as usual, but I understand that is what you want this time.

formatc1702:
Awesome, thanks!
I'll post a quick summary of my near future plans in #101.

PR #196, opened 2020-11-16
Length unit

kvid:
- This solves the basic part of #7
- It is based on #161, #162, and #171 and solves the known bugs commented in their reviews
- The commits are rebased on top of the current `dev`
- The bug about adding quantities of different units was fixed as part of the conflict resolution during rebasing
- The two commits by @zombielinux from #161 and #162 are squashed into one with authorship unchanged
- The merge commit by @stevegt disappeared during rebasing, but his real change commit is present with authorship unchanged

PR #181, opened 2020-10-17
Continuous documentation

formatc1702:
The aim of this branch is to advance in parallel to the `dev` branch as PRs are merged and features added.

This can include expanding the syntax description (`syntax.md`), the changelog (`CHANGELOG.md`), the general readme (`README.md`) or others.

Before a release, this branch can be then squashed and rebased onto `dev` to reflect all the updates included in that release.

The idea is to prevent cluttering the commit history with individual commits from each PR documenting itself (`Add #xxx to changelog`, `Expand syntax description to reflect new feature from #yyy`). It remains to be seen whether this is a good idea, and I am open to feedback on this topic.

PR #197, opened 2020-11-16
Simplify BOM code

kvid:
Currently, this PR does not change any output generated by `build_examples.py`. Consider implementing the TODO about using 'Description' in BOM heading. However, if this TODO is considered controversial or otherwise might delay merging the PR, then please ignore it for now.


formatc1702:
Nice!
Does it make sense to rebase this onto your `length_unit2` branch and include any changes necessary for it to work with the new `length_unit` attribute? (I can't thoroughly test right now, maybe my suggestion does not make sense)

I will merge #196 tonight (squashing the commits, since the overall amount of lines changed is rather small, and adding everyone as a co-author). Then this PR could seamlessly follow afterwards :)

Let me know what you think.

formatc1702:
I haven't merged this PR since I have seen some more activity on it since you removed the draft status.

If, as per [your comment](https://github.com/formatc1702/WireViz/pull/181#issuecomment-803507007), you think it is ready, I can have a look and integrate it soon :)

kvid:
> I haven't merged this PR since I have seen some more activity on it since you removed the draft status.

That's true. I've added a few commits the last months, but I guess it might be time for a review so other PRs can benefit from these changes soon (if they are accepted).

> 
> If, as per [your comment](https://github.com/formatc1702/WireViz/pull/181#issuecomment-803507007), you think it is ready, I can have a look and integrate it soon :)

The currently last commit about "Group common function arguments into a dict" might not look like a simplification, but my goal is to simplify later attribute extensions, like e.g. one of the commits in #219. The same goal is also my motivation behind the preceeding commit in this PR about `optional_fields()`.

src/wireviz/wv_bom.py
```diff
@@ -78,101 +93,89 @@ def generate_bom(harness):
...
-    for entry in harness.bom():
-        if (entry['item'], entry['unit'], entry['manufacturer'], entry['mpn'], entry['pn']) == target:
-            return entry['id']
-    return None
-
-def bom_list(bom):
-    keys = ['id', 'item', 'qty', 'unit', 'designators'] # these BOM columns will always be included
-    for fieldname in ['pn', 'manufacturer', 'mpn']: # these optional BOM columns will only be included if at least one BOM item actually uses them
+    target = tuple(clean_whitespace(v) for v in bom_types_group({**asdict(part), 'description': part.description}))
+    return next(entry['id'] for entry in bom if bom_types_group(entry) == target)
```
kvid, line 124: - Maybe we need to catch the exception and re-raise a more descriptive one?
- Alternatively,we can revert this line and use the old loop, but raise an exception instead of returning None?

formatc1702: Sorry, I have been very hands-off regarding the BOM generation code. Under what circumstances would this exception be triggered? Can it happen with the current code (where this function is only called from `﻿get_additional_component_table()` or only if it happens to be called from somewhere else?

kvid: Without this PR, None is returned if no entry found matching the search arguments. I suggest raising an exception instead of returning None to avoid such an internal error be undetected by the user. My comment here just suggest two different ways to make a more user friendly exception in such a case that normally never happen

kvid: - I have implemented the latter of the two alternatives in https://github.com/formatc1702/WireViz/pull/197/commits/9142c21e4860911cc977a114b1199e5f2442b1e4 .
- I have also added a few commits to prepare for my suggestion in https://github.com/formatc1702/WireViz/pull/226#issuecomment-805448151.

src/wireviz/wv_bom.py
```diff
@@ -78,101 +101,91 @@ def generate_bom(harness):
...
+    keys = list(BOM_COLUMNS_ALWAYS)  # Always include this fixed set of BOM columns.
+    for fieldname in BOM_COLUMNS_OPTIONAL:  # Include only those optional BOM columns that are in use.
         if any(entry.get(fieldname) for entry in bom):
             keys.append(fieldname)
-    bom_list = []
-    # list of staic bom header names,  headers not specified here are generated by capitilising the internal name
+    # Custom mapping from internal name to BOM column headers.
+    # Headers not specified here are generated by capitilising the internal name.
     bom_headings = {
+        "description": "Item",  # TODO: Remove this line to use 'Description' in BOM heading.
```
kvid, line 144: Consider removing this line and use `Description` instead of `Item` in BOM heading:
```suggestion
```
Originally, the term `item` was used both as one of the BOM column headers, and as any row in the BOM. Now, BOM `entry` is used as any row in the BOM, and `description` is used as internal name of the BOM column to be consistent with `additional_bom_items`.

If this suggestion is accepted, then this PR changes the output slightly, but only the BOM column header.

kvid: However, if this TODO is considered controversial or otherwise might delay merging the PR, then please ignore it for now.

formatc1702 (approved):
I've tested the output of this PR and it looks good.
Admittedly, I cannot currently dive deep into every single line of the code, but overall it looks like a good simplification, and I don't want to keep this open much longer than it has to. Anyways, small tweaks and fixes are always possible later.

I will squash-merge to keep the `dev` branch simple. The detailed history will be preserved in the PR branch for anyone to check.

PR #198, opened 2020-11-16
Detect and assign unit within cable length attribute

formatc1702:
Built on new `length_unit` attribute from #196.

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -200,7 +200,7 @@ class Cable:
             try:
                 g, u = self.gauge.split(' ')
             except Exception:
-                raise Exception('Gauge must be a number, or number and unit separated by a space')
+                raise Exception(f'{self.gauge} - Gauge must be a number, or number and unit separated by a space')
             self.gauge = g
 
             if u.upper() == 'AWG':
@@ -219,7 +219,7 @@ class Cable:
                 l, u = self.length.split(' ')
                 l = float(l)
             except Exception:
-                raise Exception('Length must be a number, or number and unit separated by a space')
+                raise Exception(f'{self.length} - Length must be a number, or number and unit separated by a space')
             self.length = l
             self.length_unit = u
         elif self.length is not None:  # length specified, assume m
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -214,6 +214,20 @@ class Cable:
         else:
             pass  # gauge not specified
 
+        if isinstance(self.length, str):  # length and unit specified
+            try:
+                l, u = self.length.split(' ')
+                l = float(l)
+            except Exception:
+                raise Exception('Length must be a number, or number and unit separated by a space')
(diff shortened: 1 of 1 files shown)


src/wireviz/DataClasses.py
```diff
@@ -214,6 +214,20 @@ def __post_init__(self) -> None:
...
+                l = float(l)
+            except Exception:
+                raise Exception(f'{self.length} - Length must be a number, or number and unit separated by a space')
+            self.length = l
+            self.length_unit = u
+        elif self.length is not None:  # length specified, assume m
+            if self.gauge_unit is None:
+                self.length_unit = 'm'
+        else:
+            pass  # length not specified
```
kvid, line 229: As `length` has `0` as default value, it can only get `None` if specified as `null`, `Null`, `NULL`, or `~` in the input YAML, and that should probably raise an error exception.
```suggestion
        elif not any(isinstance(self.length, t) for t in [int, float]):
            raise Exception(f'Cable {self.name} length has a non-numeric value')
        elif self.gauge_unit is None:
            self.length_unit = 'm'
```
Assigning the default unit is also done below, and is not needed twice.

src/wireviz/DataClasses.py
```diff
@@ -214,6 +214,20 @@ def __post_init__(self) -> None:
...
+
         self.connections = []
 
         if self.length_unit is None: #Default wire length units to meters if left undeclared
```
kvid, line 233: This `if` and the belonging next line (github don't allow me to comment code too far off your changes) is not needed twice. In the new code above it is done only when needed.

src/wireviz/DataClasses.py
```diff
@@ -200,7 +200,7 @@ def __post_init__(self) -> None:
             try:
                 g, u = self.gauge.split(' ')
             except Exception:
-                raise Exception('Gauge must be a number, or number and unit separated by a space')
+                raise Exception(f'{self.gauge} - Gauge must be a number, or number and unit separated by a space')
```
kvid, line 203:
```suggestion
                raise Exception(f'Cable {self.name} gauge={self.gauge} - Gauge must be a number, or number and unit separated by a space')
```

src/wireviz/DataClasses.py
```diff
@@ -214,6 +214,20 @@ def __post_init__(self) -> None:
         else:
             pass  # gauge not specified
 
+        if isinstance(self.length, str):  # length and unit specified
+            try:
+                l, u = self.length.split(' ')
+                l = float(l)
+            except Exception:
+                raise Exception(f'{self.length} - Length must be a number, or number and unit separated by a space')
```
kvid, line 222:
```suggestion
                raise Exception(f'Cable {self.name} length={self.length} - Length must be a number, or number and unit separated by a space')
```

kvid (changes requested):

formatc1702:
Thanks for the suggestions, they've all been included.

kvid (approved):

PR #74, opened 2020-07-09
Improve output of diagram + BOM as technical drawing

formatc1702:
Closes #32.

PR #214, opened 2021-02-09
Add basic options and metadata

kvid:
- Add metadata attributes: title, description, notes
- Add option attributes: color_mode, mini_bom_mode
- Use metadata values and a few more indents and linebreaks in the HTML output
- Translate color text according to options.color_mode a few more places
- This solves the initial part of #158

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -20,6 +20,7 @@ ConnectorMultiplier = PlainText # = Literal['pincount', 'populated']
 CableMultiplier = PlainText # = Literal['wirecount', 'terminations', 'length', 'total_length']
 ImageScale = PlainText # = Literal['false', 'true', 'width', 'height', 'both']
 Color = PlainText # Two-letter color name = Literal[wv_colors._color_hex.keys()]
+ColorMode = PlainText # = Literal['full', 'FULL', 'hex', 'HEX', 'short', 'SHORT', 'ger', 'GER']
 ColorScheme = PlainText # Color scheme name = Literal[wv_colors.COLOR_CODES.keys()]
 
 # Type combinations
@@ -28,6 +29,19 @@ Pin = Union[int, PlainText] # Pin identifier
 Wire = Union[int, PlainText] # Wire number or Literal['s'] for shield
 NoneOrMorePins = Union[Pin, Tuple[Pin, ...], None] # None, one, or a tuple of pins
 OneOrMoreWires = Union[Wire, Tuple[Wire, ...]] # One or a tuple of wires
+
+
+@dataclass
+class Metadata:
+    title: PlainText
+    description: Optional[MultilineHypertext] = None
+    notes: Optional[MultilineHypertext] = None
+
+
+@dataclass
+class Options:
+    color_mode: ColorMode = 'SHORT'
+    mini_bom_mode: bool = True
 
 
 @dataclass
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -4,12 +4,13 @@
 from graphviz import Graph
 from collections import Counter
 from typing import List, Union
+from dataclasses import dataclass
 from pathlib import Path
 from itertools import zip_longest
 import re
 
 from wireviz import wv_colors, __version__, APP_NAME, APP_URL
-from wireviz.DataClasses import Connector, Cable
+from wireviz.DataClasses import Metadata, Options, Connector, Cable
 from wireviz.wv_colors import get_color_hex
 from wireviz.wv_gv_html import nested_html_table, html_colorbar, html_image, \
     html_caption, remove_links, html_line_breaks
@@ -20,11 +21,12 @@ from wireviz.wv_helper import awg_equiv, mm2_equiv, tuplelist2tsv, flatten2d, \
     open_file_read, open_file_write
 
 
+@dataclass
 class Harness:
+    metadata: Metadata
+    options: Options
 
-    def __init__(self):
-        self.color_mode = 'SHORT'
-        self.mini_bom_mode = True
+    def __post_init__(self):
         self.connectors = {}
         self.cables = {}
         self._bom = []  # Internal Cache for generated bom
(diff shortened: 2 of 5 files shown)


formatc1702:
Does it make sense to keep the `metadata` section more general, as a dictionary to be filled with user-defined key/value pairs instead of limiting it to certain fixed parameters?
Certain keywords could have effects in the GraphViz output (`title` maybe?), whereas others could be accessed from HTML when [outputting a full technical drawing](https://github.com/formatc1702/WireViz/issues/158#issuecomment-711065152) and allow for expansion with parameters that users will [think of in the future](https://github.com/formatc1702/WireViz/pull/74#issuecomment-802714860).

kvid:
> Does it make sense to keep the `metadata` section more general, as a dictionary to be filled with user-defined key/value pairs instead of limiting it to certain fixed parameters?

If you by this mean replacing the Metadata dataclass with a dict, then that is quite easy to do. The drawback is more complex syntax when using the values in the code.

> Certain keywords could have effects in the GraphViz output (`title` maybe?),

The currently suggested code in this PR only use metadata for generating HTML.

> whereas others could be accessed from HTML when [outputting a full technical drawing](https://github.com/formatc1702/WireViz/issues/158#issuecomment-711065152) and allow for expansion with parameters that users will [think of in the future](https://github.com/formatc1702/WireViz/pull/74#issuecomment-802714860) for use with their own templates.

formatc1702:
> > Does it make sense to keep the `metadata` section more general, as a dictionary to be filled with user-defined key/value pairs instead of limiting it to certain fixed parameters?
> 
> If you by this mean replacing the Metadata dataclass with a dict, then that is quite easy to do. The drawback is more complex syntax when using the values in the code.

This may be a rare instance where it makes sense to [implement dot notation for a dict](https://dev.to/0xbf/use-dot-syntax-to-access-dictionary-key-python-tips-10ec) then! By creating a custom class like that, you could force it to have the standard attributes (`title`, `description`, `notes` in your suggested code) and auto-fill them with default values in the class' `__init__()` if the YAML input does not provide them, and at the same time, leave it flexible and open to user-defined attributes they might need in their drawing template.

Maybe this is overkill, but it could be a good compromise.

> > Certain keywords could have effects in the GraphViz output (`title` maybe?),
> 
> The currently suggested code in this PR only use metadata for generating HTML.

Just leaving this as an idea for the future.

src/wireviz/Harness.py
```diff
@@ -20,11 +21,12 @@
     open_file_read, open_file_write
 
 
+@dataclass
 class Harness:
+    metadata: Metadata
```
formatc1702, line 26:
```suggestion
    metadata: Dict
```

src/wireviz/wireviz.py
```diff
@@ -34,7 +35,11 @@ def parse(yaml_input: str, file_out: (str, Path) = None, return_types: (None, st
 
     yaml_data = yaml.safe_load(yaml_input)
 
-    harness = Harness()
+    # Assign default metadata.title here to avoid needing file_out in Metadata.__post_init__().
+    harness = Harness(
+        Metadata(**{'title': Path(file_out).stem, **yaml_data.get('metadata', {})}),
+        Options(**yaml_data.get('options', {})),
+    )
```
formatc1702, line 42:
```suggestion
    harness = Harness(
        metadata = yaml_data.get('metadata', {}),
        options = Options(**yaml_data.get('options', {})),
    )

    if not 'title' in harness.metadata:
        harness.metadata['title'] = Path(file_out).stem
```

A more pythonic and readable suggestion, assuming `Metadata` is kept as a normal `Dict`

src/wireviz/DataClasses.py
```diff
@@ -30,6 +31,22 @@
 OneOrMoreWires = Union[Wire, Tuple[Wire, ...]] # One or a tuple of wires
 
 
+@dataclass
+class Metadata:
+    title: PlainText
+    description: Optional[MultilineHypertext] = None
+    notes: Optional[MultilineHypertext] = None
+
```
formatc1702, line 39:
```suggestion

```

src/wireviz/Harness.py
```diff
@@ -4,13 +4,14 @@
...
 from typing import List, Union
+from dataclasses import dataclass
 from pathlib import Path
 from itertools import zip_longest
 import re
 
 from wireviz import wv_colors, __version__, APP_NAME, APP_URL
-from wireviz.DataClasses import Connector, Cable
-from wireviz.wv_colors import get_color_hex
+from wireviz.DataClasses import Metadata, Options, Connector, Cable
```
formatc1702, line 13:
```suggestion
from wireviz.DataClasses import Options, Connector, Cable
```

src/wireviz/wireviz.py
```diff
@@ -13,6 +13,7 @@
     sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
 
 from wireviz import __version__
+from wireviz.DataClasses import Metadata, Options
```
formatc1702, line 16:
```suggestion
from wireviz.DataClasses import Options
```

src/wireviz/wv_html.py
```diff
@@ -2,21 +2,27 @@
...
+from typing import List, Union
 import re
 
-from wireviz import __version__, APP_NAME, APP_URL
+from wireviz import __version__, APP_NAME, APP_URL, wv_colors
+from wireviz.DataClasses import Metadata, Options
 from wireviz.wv_helper import flatten2d, open_file_read, open_file_write
 
-def generate_html_output(filename: (str, Path), bom_list):
+def generate_html_output(filename: Union[str, Path], bom_list: List[List[str]], metadata: Metadata, options: Options):
```
formatc1702, line 12: I haven't included any suggestions here, since this function will hopefully be rebuilt in #74 soon.

src/wireviz/Harness.py
```diff
@@ -4,13 +4,14 @@
 from graphviz import Graph
 from collections import Counter
 from typing import List, Union
```
formatc1702, line 6:
```suggestion
from typing import List, Dict, Union
```

src/wireviz/Harness.py
kvid: I suggest defining `Metadata = Dict` in `DataClasses.py` to make it easy to expand it to a more advanced type later if needed, and to enable continue using `Metadata` as argument type hints, e.g. in `generate_html_output()`.

formatc1702: OK! That sounds like a good compromise.

formatc1702:
I have tried out the code, and it works well. I also have no issues with the code style itself.

Would you mind adding the required documentation to `syntax.md`? It's the only thing missing for merging.
Thanks!

Issue #174, opened 2020-10-12
[feature] Force increased spacing and ordering for readability


tim292stro:
So I'm actually trying to use this to describe a cable rework at my day job - and I have a doozy of a cable to describe:

{START}
connectors:
  CableEnd_Near:
    type: HDMI
    subtype: male
    pinout: [ TMDS_D2+, TMDS_D2Shield, TMDS_D2-, TMDS_D1+, TMDS_D1Shield, TMDS_D1-, TMDS_D0+, TMDS_D0Shield, TMDS_D0-, TMDS_CLK+, TMDS_CLKShield, TMDS_CLK-, CEC, HEAC+, DDC_SCL, DDC_SDA, DDC/CEC/HEAC_Ground, +5V, HPD/HEAC-, SHELL]
  CableEnd_Far:
    type: HDMI
    subtype: male
    pinout: [ TMDS_D2+, TMDS_D2Shield, TMDS_D2-, TMDS_D1+, TMDS_D1Shield, TMDS_D1-, TMDS_D0+, TMDS_D0Shield, TMDS_D0-, TMDS_CLK+, TMDS_CLKShield, TMDS_CLK-, CEC, HEAC+, DDC_SCL, DDC_SDA, DDC/CEC/HEAC_Ground, +5V, HPD/HEAC-, SHELL]

cables:
  TPWhiteSilver:
    wirecount: 2
    length: 910
    shield: true
    colors: [WH, BK]
  TPYellowSilver:
    wirecount: 2
    length: 910
    shield: true
    colors: [WH, BK]
  TPGreenSilver:
    wirecount: 2
    length: 910
    shield: true
    colors: [WH, BK]
  TPPinkSilver:
    wirecount: 2
    length: 910
    shield: true
    colors: [WH, BK]
  TPBlueSilver:
    wirecount: 2
    length: 910
    shield: true
    colors: [WH, BK]
  Sideband:
    wirecount: 4
    length: 910
    shield: true
    colors: [RD, BU, YE, WH]

connections:
  -
    - CableEnd_Near: [1-3]
    - TPWhiteSilver: [1,s,2]
    - CableEnd_Far: [1-3]
  -
    - CableEnd_Near: [4-6]
    - TPYellowSilver: [1,s,2]
    - CableEnd_Far: [4-6]
  -
    - CableEnd_Near: [7-9]
    - TPGreenSilver: [1,s,2]
    - CableEnd_Far: [7-9]
  -
[...]

PR #215, opened 2021-02-11
Add optional tweaking of the .gv output

kvid:
- Solves #174 intermediately by allowing low level .gv tweaking.
- See https://github.com/formatc1702/WireViz/issues/174#issuecomment-707372543 and [HDMI_Sideband_Tap_tweaked.txt](https://github.com/formatc1702/WireViz/files/7130401/HDMI_Sideband_Tap_tweaked.txt) for an example usage and suggested changes of `syntax.md` for accepted input value types.
- This tweak feature is not limited to what I suggest in that example usage. It enables overriding any main attribute of any entry in the generated `.gv` file, and appending any number of custom entries, e.g. nodes, edges, and subgraphs. Overriding entries with attributes containing HTML is not tested, and will probably need more advanced functionality, but no such plans without any requests.
- Please check that attributes and custom entries are legal according to [The DOT Language](https://graphviz.org/doc/info/lang.html). Otherwise, Graphviz will issue error messages.
- ~~This PR also contains the commits from #214 and will probably need a rebase after #214 gets merged in.~~ (rebase done)
- Please try this feature by cloning my https://github.com/kvid/WireViz/tree/issue174-tweak branch and write a comment in this PR about your experience, but be aware that until this PR is merged into `dev`, I might change the contents of my branch at any time, including force-push that breaks old commit history.

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -345,10 +345,11 @@ Alternatively items can be added to just the BOM by putting them in the section 
     # Entries with an attribute containing HTML are
     # not supported.
     <str>:  # leading string of .gv entry
-      <str> : <str>  # attribute and its new value
+      <str> : <str/null>  # attribute and its new value
       # Any number of attributes can be overridden
       # for each entry. Attributes not already existing
       # in the entry will be appended to the entry.
+      # Use null as new value to delete an attribute.
 
   append: <str/list> # string or list of strings to append to the .gv output
 ```
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -35,6 +35,8 @@ additional_bom_items:  # custom items to add to BOM
   - <bom-item>           # BOM item (see below)
   ...
 
+tweak:  # optional tweaking of .gv output
+  ...
 ```
 
 ## Metadata entries
@@ -327,6 +329,30 @@ Alternatively items can be added to just the BOM by putting them in the section 
   manufacturer: <str>  # manufacturer name  
 ```
 
+## Tweak entries
+
+```yaml
+  # Optional tweaking of the .gv output.
+  # This feature is experimental and might change
+  # or be removed in future versions.
+
+  override:  # dict of .gv entries to override
+    # Each entry is identified by its leading string
+    # in lines beginning with a TAB character.
+    # The leading string might be in "quotes" in 
+    # the .gv output. This leading string must be
+    # followed by attributes in [square brackets].
+    # Entries containing HTML in an attribute are
+    # not supported.
+    <str>:  # leading string of .gv entry
+      <str> : <str>  # attribute and its new value
+      # Any number of attributes can be overridden
+      # for each entry
+
+  append:  # single or list of .gv entries to append
+    <str>  # strings to append might have multiple lines
+```
+
 ## Colors
 
 Colors are defined via uppercase, two character strings.
@@ -403,6 +429,7 @@ The following attributes accept multiline strings:
 - `manufacturer`
 - `mpn`
 - `image.caption`
+- `tweak.append`
 
 ### Method 1
 
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -347,7 +347,8 @@ Alternatively items can be added to just the BOM by putting them in the section 
     <str>:  # leading string of .gv entry
       <str> : <str>  # attribute and its new value
       # Any number of attributes can be overridden
-      # for each entry
+      # for each entry. Attributes not already existing
+      # in the entry will be appended to the entry.
 
   append: <str/list> # string or list of strings to append to the .gv output
 ```
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -329,7 +329,7 @@ Alternatively items can be added to just the BOM by putting them in the section 
   manufacturer: <str>  # manufacturer name  
 ```
 
-## Tweak entries
+## GraphViz tweaking (experimental)
 
 ```yaml
   # Optional tweaking of the .gv output.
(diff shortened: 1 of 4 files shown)


tim292stro:
Will check it this weekend

tim292stro:
Man that was a long "this weekend" finally getting to it this afternoon... :-|

formatc1702:
Could you please provide a sample file where you `append` and `override` some stuff in the YAML?
Doesn't have to be part of the PR, a comment here is enough.
Thanks!

docs/syntax.md
```diff
@@ -327,6 +329,30 @@ Alternatively items can be added to just the BOM by putting them in the section
...
+    # followed by attributes in [square brackets].
+    # Entries containing HTML in an attribute are
+    # not supported.
+    <str>:  # leading string of .gv entry
+      <str> : <str>  # attribute and its new value
+      # Any number of attributes can be overridden
+      # for each entry
+
+  append:  # single or list of .gv entries to append
+    <str>  # strings to append might have multiple lines
```
formatc1702, line 353:
```suggestion
  append: <str/list> # string or list of strings to append to the .gv file
```

src/wireviz/Harness.py
```diff
@@ -344,6 +345,44 @@ def create_graph(self) -> Graph:
...
+                if isinstance(entry, str):
+                    # Find a possibly quoted keyword after leading TAB(s) and followed by [ ].
+                    match = re.match(r'^\t*(")?((?(1)[^"]|[^ "])+)(?(1)") \[.*\]$', entry, re.S)
+                    keyword = match and match[2]
+                    if keyword in self.tweak.override.keys():
+                        for attr, value in self.tweak.override[keyword].items():
+                            if len(value) == 0 or ' ' in value:
+                                value = value.replace('"', r'\"')
+                                value = f'"{value}"'
+                            # TODO?: If value is None: delete attr, and if attr not found: append it?
```
formatc1702, line 373: both ideas make sense

docs/syntax.md
```diff
@@ -327,6 +329,30 @@ Alternatively items can be added to just the BOM by putting them in the section
   manufacturer: <str>  # manufacturer name  
 ```
 
+## Tweak entries
```
formatc1702, line 332:
```suggestion
## GraphViz tweaking (experimental)
```

kvid:
> Could you please provide a sample file where you `append` and `override` some stuff in the YAML?
> Doesn't have to be part of the PR, a comment here is enough.
> Thanks!

See the already provided link to such an example in the second bullet of this PR description.

src/wireviz/Harness.py
```diff
@@ -344,6 +345,44 @@ def create_graph(self) -> Graph:
             dot.node(cable.name, label=f'<\n{html}\n>', shape='box',
                      style=style, fillcolor=translate_color(bgcolor, "HEX"))
 
+        def typecheck(name: str, var, type) -> None:
+            if not isinstance(var, type):
+                raise Exception(f'Unexpected value type of {name}: {var}')
```
formatc1702, line 350:
```suggestion
        def typecheck(name: str, var, typ) -> None:
            if not isinstance(var, typ):
                raise Exception(f'Unexpected value type of {name}: Expected {typ}, got {type(var)}\n{var}')
```

Renaming `type` to `typ` to avoid using the name of the built-in `type()` function.

src/wireviz/Harness.py
```diff
@@ -344,6 +345,44 @@ def create_graph(self) -> Graph:
...
+                        for attr, value in self.tweak.override[keyword].items():
+                            if len(value) == 0 or ' ' in value:
+                                value = value.replace('"', r'\"')
+                                value = f'"{value}"'
+                            # TODO?: If value is None: delete attr, and if attr not found: append it?
+                            entry = re.sub(f'{attr}=("[^"]*"|[^] ]*)', f'{attr}={value}', entry)
+                        dot.body[i] = entry
+
+        if self.tweak.append is not None:
+            if isinstance(self.tweak.append, list):
```
formatc1702, line 378: Is there any reason that the `append` section should be a list, and not a string?
Is there a benefit to having multiple list entries for different tweaks, as opposed to all of them concatenated into one string? I am not opposed to it, just genuinely curious.

src/wireviz/Harness.py
formatc1702:
```suggestion
        def typecheck(name: str, var, type_) -> None:
            if not isinstance(var, type_):
                raise Exception(f'Unexpected value type of {name}: Expected {type_}, got {type(var)}\n{var}')
```

Apparently, [PEP 8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) recommends single trailing underscore for these situations.

src/wireviz/Harness.py
kvid: Initially, I used a list of quoted strings to easily specify leading spaces where I wanted them. Leading spaces in multiline strings seems to be stripped by the YAML parser, but maybe you know a way to avoid that?

src/wireviz/Harness.py
kvid: Yes, that was the plan. The current implementation should be enough to solve the original issue and a lot of similar experimental tweaking, but I don't expect a huge number of users will need such tweaking in their harnesses. Maybe it'll mainly be a tool for developers to test variations of `.gv` output before suggesting a proper implementation of a new feature.

src/wireviz/Harness.py
formatc1702: If you include the indentation indicator ([source](https://yaml-multiline.info/)), any additional leading spaces are preserved. I quickly tested this and it works. Example:

```yaml
tweak:
  append: |2  # two leading spaces will be interpreted as indentation; any more will be preserved
              // Tweaking two cable nodes to keep close to each other
              subgraph cluster1 {
                color=yellow // Same as background to hide the surrounding frame
                TPPinkSilver
                TPBlueSilver
              }
```

src/wireviz/Harness.py
kvid: I just implemented these features as separate commits in this PR as well.

formatc1702 (approved):

formatc1702:
Thanks. I'll leave #174 open, since it is not directly solved by this. If there is any other issue/PR you think should be included in the [changelog](https://github.com/formatc1702/WireViz/commit/4da115489ca333571325bb2d3f8f3e98732ce1f8), let me know!

PR #221, opened 2021-03-09
Correctly check for show_wirenumbers default of None

gopiballava:
If `show_wirenumbers` is omitted from a cable section, its value will be `None`. In that case, we want to choose a default based on whether this is a bundle or not.

If the user did specify `show_wirenumbers`, then its value will be `True` or `False`, and we want to respect that whether it's a bundle or not.

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -269,7 +269,7 @@ class Cable:
                     raise Exception('lists of part data are only supported for bundles')
 
         # by default, show wire numbers for cables, hide for bundles
-        if not self.show_wirenumbers:
+        if self.show_wirenumbers is None:
             self.show_wirenumbers = self.category != 'bundle'
 
         for i, item in enumerate(self.additional_components):


kvid (approved):
I recommend accepting this PR based on a simple test with this YAML input:
```yaml
cables:
  MyCable:
    colors: ['BN', 'RD', 'YE']
    show_wirenumbers: False
```
The left diagram shows that the result without this PR doesn't respect the `show_wirenumbers: False` for non-bundle cables, and the right diagram shows the result from the same YAML input after applying this PR:
issue221-before issue221-after
I have also verified that none of the generated example files change after applying this PR:
```bash
git checkout -b dev-built origin/dev
python build_examples.py
git add ../../{examples,tutorial}/*
git commit
python build_examples.py restore
git checkout gopiballava-fix-show_wirenumbers
python build_examples.py
python build_examples.py -b dev-built -c compare
python build_examples.py restore
```

Issue #233, opened 2021-06-07
[feature] Embed YAML data into PNG


jacobian91:
PNGs have the ability to add arbitrary data to them in 'chunks'. This is inspired by diagram software [diagrams.net](https://www.diagrams.net/blog/xml-in-png). The XML that creates the diagram is stored in the zTXT chunk of the PNG. For WireViz, the equivalent would be putting the YAML data.

In addition to storing the data inside, it is most useful if WireViz can then use it as the input file, and ideally output the YAML file as well.

formatc1702:
Hi,
sorry for ignoring this issue and the associated PR for this long.
This is an interesting idea... can you explain a real-life situation where this might come in handy?

It's nice that the PNG is "self-documenting", but I'm wondering in which situation you would want to use a PNG as an input file directly?

jacobian91:
This is especially nice for document control systems in companies that produce hardware. Those systems typically require 2 files to be uploaded for something like this, rendered and source. That way the parties who need each can access them in a single controlled location. However, every time there is an update both of the files need to be kept in sync. Doc-control systems are notoriously manual and so files getting out of sync is a matter of human error. Helping prevent this issue reduces time/maintenance costs for companies who use it.

kvid:
The "YAML data" might be the sum of a potentially large prepended file followed by the main input file. What should then be stored in the PNG file? If storing a copy of the prepended file, then it'll be disconnected from later changes to a prepended file that might be common to many different main YAML files. Is that a wanted feature?

jacobian91:
@kvid I had originally kept only the complete yaml because of the structure of the code and I couldn't separate the prepend from the primary yaml file. I added some arguments to the main parse function now that give that ability so that the png now contains the yaml data file and the prepend yaml file, and when reading from a png file both will be saved to the directory as `_out` files.

This way it is easy to ignore the prepend file and still recover the diagram-specific yaml. Thanks for the suggestion!

kvid:
@jacobian91 I like your solution to my scenario above. Splitting the prepended and main YAML into separate embedded sections will indeed make it easier to restore the original YAML files - even in cases with a common prepended file shared with multiple main files.

[...]

Issue #190, opened 2020-10-24
[internal] Replace`os.path` with `Pathlib`


formatc1702:
Remove all references to `os.path` in the code and replace with the equivalents using `Pathlib` for consistency and robustness.

- [Source 1.1](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/)
- [Source 1.2](https://treyhunner.com/2019/01/no-really-pathlib-is-great/)

formatc1702:
Perhaps @n42 can explain the need for adding the `..` in [line 13 in `wireviz.py`](https://github.com/formatc1702/WireViz/blame/master/src/wireviz/wireviz.py#L13)? Isn't it enough to insert `os.path.dirname(__file__)` to find the WireViz modules?

Would the `Pathlib` equivalent then be `Path(__file__).parent`? Thanks :)

PR #234, opened 2021-06-07
Embed YAML data into PNG

jacobian91:
Is a solution for #233.

- Saves YAML data to zTXT chunk of PNGs (note: It does not distinguish between prepended file and input file)
- Allows WireViz to use a PNG as the input file.
- If the input file is a PNG, a YAML file is created as one of the outputs. Which can be run again by standard WireViz.


src/wireviz/wireviz.py
```diff
@@ -232,16 +245,21 @@ def main():
...
-        yaml_input = fh.read()
-
-    if args.prepend_file:
-        if not os.path.exists(args.prepend_file):
-            print(f'Error: prepend input file {args.prepend_file} inaccessible or does not exist, check path')
-            sys.exit(1)
-        with open_file_read(args.prepend_file) as fh:
-            prepend = fh.read()
-            yaml_input = prepend + yaml_input
+    if ".png" in args.input_file:
```
formatc1702, line 248: Please consider using `Pathlib`'s `.suffix` attribute for checking the file extension.

src/wireviz/wireviz.py
```diff
@@ -232,16 +245,21 @@ def main():
...
+        yaml_input = read_yaml_from_png(args.input_file.replace('.png',''))
+        with open(args.input_file.replace('.png','_out.yaml'),'w') as fh:
```
formatc1702, line 250: Consider using `wv_helper.py`'s `﻿﻿open_file_read()` and `open_file_write()` functions for file access to ensure UTF-8 encoding

src/wireviz/wireviz.py
```diff
@@ -232,16 +245,21 @@ def main():
...
+            fh.write(yaml_input) # Extract yaml to separate file
+    else:
+        with open_file_read(args.input_file) as fh:
+            yaml_input = fh.read()
+
+        if args.prepend_file:
+            if not os.path.exists(args.prepend_file):
```
formatc1702, line 257: Consider using `Pathlib`'s `exists()` function instead (see #190)

src/wireviz/wireviz.py
```diff
@@ -232,16 +245,21 @@ def main():
...
+                print(f'Error: prepend input file {args.prepend_file} inaccessible or does not exist, check path')
+                sys.exit(1)
```
formatc1702, line 259: Please raise an exception instead, for consistency with the rest of the code

formatc1702 (commented):
Thanks for your explanation in #233.
I will play around with and think about this feature a bit later.
For now, I'm leaving some comments on the current implementation.
I will come back to this PR once I have time.

src/wireviz/wireviz.py
jacobian91: Github is reporting this as a change, but this is mostly whitespace changes only, I only indented here.

formatc1702: true!

src/wireviz/wireviz.py
formatc1702: true. will fix separately.

src/wireviz/wireviz.py
jacobian91: Done, thanks!

src/wireviz/wireviz.py
jacobian91: Done

jacobian91:
I changed the `argparse` types to come in as `pathlib.Path` object types to shortcut some of the conversion that would have to otherwise be done in each of the functions.

Tyler-Ward:
A thought with regards to the implementation it might be worth either putting this capability behind a command line option, adding a command line option to not add the yaml data into the png (depending on the preference for weither to enable by default or not) or otherwise having a way to get an ouptut without the input embedded.

The reasons I can think of for not wanting to add the input into the png are.
- Keeping othersise unexpected/sensitive data out of a diagram output e.g. data from the prepend not used in the diagram or in comments.
- Avoiding the png file size increase, although probably not a significant concern unless a long prepend file is used.

Another soloution for both of these would be to trim down/regenerate the yaml added to the png to only that which is needed to generate the diagram but that feels like it could be quite a complicated thing to do.

jacobian91:
@Tyler-Ward, really good points, I hadn't thought of those aspects.

> Avoiding the png file size increase, although probably not a significant concern unless a long prepend file is used.

A good thing to check, I just ran it with a 4KB prepend & 3.2KB diagram file. The ending png is 213KB with the data, or 211KB without the data. The size difference I'm assuming is the compressed state of the yaml, which results from the line: `txt.add_itxt('yaml', yaml_input,zip=True`. 

> Keeping otherwise unexpected/sensitive data out of a diagram output e.g. data from the prepend not used in the diagram or in comments.

I think that is a good point, I'll leave that up to @formatc1702 for comment. I can certainly implement that, but the questions would be: 

1. Do you want to allow 'source' to be distributed or not, and are the licensing preferences that come into play here?
2. If you do want the option, what would you like the default behavior to be? I'm in favor of default being to share the source. I've lost source files before and having this be an automatic backup is certainly helpful.

jacobian91:
The other formats, SVG & HTML also could incorporate these but those implementations are a bit more straightforward because they could just be added as comments if desired, and would be able to be just copy/pasted instead of needing to be read by the library. But since those will not have a compression option built-in, the file size will more likely be 1:1 increase by adding the data.

jacobian91:
Update: Rebased on the latest dev version for easier merging

jacobian91:
@Tyler-Ward I just added a command line argument to 'conceal-input' so that this feature can be turned off.

PR #241, opened 2021-08-27
Implement suppliers and supplier part numbers

### LINKED ISSUE #240: [feature] Add supplier / SPN info
This has arisen as a personal need.

The current component attributes `manufacturer` and `mpn` are great for generating useful BOMs.
However, often it is also important to specify a specific supplier, which likely uses an independent part numbering system.

Proposal: add `supplier` and `spn` fields, analogously to `manufacturer` and `mpn`.

Obviously, it may also be desired to provide more than one supplier for a given part. That feature is beyond the scope of this feature proposal; realistically, complex supplier management should be left to a dedicated ERP system, which could map company-internal part numbers (using the `pn` attribute) to any number of manufacturers and suppliers.

formatc1702:
Closes #240.


tutorial/tutorial08.yml
```diff
@@ -35,6 +50,8 @@ cables:
     colors: [YE, BK, BK, RD]
     manufacturer: [WiresCo,WiresCo,WiresCo,WiresCo] # set a manufacter per wire
     mpn: [W1-YE,W1-BK,W1-BK,W1-RD]
+    supplier: [WireShack,WireShack,WireShack,WireShack]
```
kvid, line 53: When all wires in a bundle have the same attribute value, it's also possible to specify this attribute value as a single string. In that case, the common value will be shown in the top part of the node, though. The BOM should be the same in both cases.

src/wireviz/wv_bom.py
```diff
@@ -156,26 +157,28 @@ def component_table_entry(
...
         manufacturer: Optional[str] = None,
         mpn: Optional[str] = None,
+        supplier: Optional[str] = None,
+        spn: Optional[str] = None,
     ) -> str:
     """Return a diagram node table row string with an additional component."""
-    manufacturer_str = manufacturer_info_field(manufacturer, mpn)
+    manufacturer_str = pn_info_string("mpn", manufacturer, mpn)
+    supplier_str = pn_info_string("spn", supplier, spn)
+    part_number_list = [pn, manufacturer_str, supplier_str]
```
kvid, line 166: Consider alternative:
```suggestion
    part_number_list = [pn, pn_info_string(HDR_MPN, manufacturer, mpn), pn_info_string(HDR_SPN, supplier, spn)]
```
or maybe:
```suggestion
    part_number_list = [
        pn,
        pn_info_string(HDR_MPN, manufacturer, mpn),
        pn_info_string(HDR_SPN, supplier, spn),
    ]
```
unless readability is hurt.

kvid (changes requested):
Except for one bug description, this review mainly suggests you consider a few code alternatives with no change of output.

src/wireviz/wv_bom.py
```diff
@@ -156,26 +157,28 @@ def component_table_entry(
...
     output = (f'{qty}'
               + (f' {unit}' if unit else '')
               + f' x {type}'
-              + ('<br/>' if pn or manufacturer_str else '')
-              + (f'P/N: {pn}' if pn else '')
```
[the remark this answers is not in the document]
formatc1702: accidental, fixed!

tutorial/tutorial08.yml
kvid: Should we offer an option to always show such a common attribute value for each wire in a bundle?

kvid (changes requested):
Thank's for implementing most of my suggestions in my previous review. However, there are a few issues that I forgot to mention.

tutorial/tutorial08.yml
formatc1702: I think the current way is fine.
If the attribute is a string, display it once in the same row as `length` etc.
If it is a list, display one list item per wire.

src/wireviz/wv_bom.py
```diff
@@ -156,26 +161,30 @@ def component_table_entry(
...
   </tr></table>'''
 
-def manufacturer_info_field(manufacturer: Optional[str], mpn: Optional[str]) -> Optional[str]:
-    """Return the manufacturer and/or the mpn in one single string or None otherwise."""
-    if manufacturer or mpn:
-        return f'{manufacturer if manufacturer else "MPN"}{": " + str(mpn) if mpn else ""}'
+def pn_info_string(nametype: str, name: Optional[str], number: Optional[str]) -> Optional[str]:
+    """Return the company name and/or the part number in one single string or None otherwise."""
+    if name or number:
+        return f'{name if name else nametype.upper()}{": " + str(number) if number else ""}'
```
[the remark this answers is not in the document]
formatc1702: - Removed `upper()`
- I am not a fan of `x or y` when working with strings; I understand why it works but I prefer `x if x else y`.. more intuitive
- Will restructure this function and test it against 0 and others

kvid (approved):

PR #246, opened 2021-10-07
Replace `os.path` with `pathlib.Path`

### LINKED ISSUE #190: [internal] Replace`os.path` with `Pathlib`

formatc1702:
Closes #190.

### DIFF
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -9,7 +9,7 @@ from typing import Any, Tuple
 import yaml
 
 if __name__ == '__main__':
-    sys.path.insert(0, str(Path(__file__).parent.parent))
+    sys.path.insert(0, str(Path(__file__).parent.parent))  # add src/wireviz to PATH
 
 from wireviz import __version__
 from wireviz.DataClasses import Metadata, Options, Tweak
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -312,11 +312,13 @@ def main():
 
     args = parse_cmdline()
 
-    if not Path(args.input_file).exists():
-        print(f'Error: input file {args.input_file} inaccessible or does not exist, check path')
+    file_in = Path(args.input_file)
+
+    if not file_in.exists():
+        print(f'Error: input file {file_in} inaccessible or does not exist, check path')
         sys.exit(1)
 
-    with open_file_read(args.input_file) as fh:
+    with open_file_read(file_in) as fh:
         yaml_input = fh.read()
 
     if args.prepend_file:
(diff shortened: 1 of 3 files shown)


src/wireviz/wireviz.py
```diff
@@ -311,28 +312,28 @@ def main():
...
             sys.exit(1)
         with open_file_read(args.prepend_file) as fh:
             prepend = fh.read()
             yaml_input = prepend + yaml_input
 
     if not args.output_file:
-        file_out = args.input_file
-        pre, _ = os.path.splitext(file_out)
-        file_out = pre  # extension will be added by graphviz output function
+        file_out = file_in.parent / file_in.stem  # extension will be added by graphviz output function
```
kvid, line 333: This is a simpler and probably more readable equivalent:
```suggestion
        file_out = file_in.with_suffix('')  # Suffix will be added by graphviz output function
```

src/wireviz/wireviz.py
```diff
@@ -311,28 +312,28 @@ def main():
...
     if args.prepend_file:
-        if not os.path.exists(args.prepend_file):
+        if not Path(args.prepend_file).exists():
             print(f'Error: prepend input file {args.prepend_file} inaccessible or does not exist, check path')
             sys.exit(1)
         with open_file_read(args.prepend_file) as fh:
             prepend = fh.read()
             yaml_input = prepend + yaml_input
 
     if not args.output_file:
```
kvid, line 332: Consider swapping the `if` and the `else` statements and remove the `not`:
```suggestion
    if args.output_file:
```

kvid (changes requested):
I also like the pathlib interface. Consider the change suggested by @laurierloi and a few other code alternatives.

src/wireviz/wireviz.py
```diff
@@ -311,28 +312,28 @@ def main():
...
+        print(f'Error: input file {file_in} inaccessible or does not exist, check path')
         sys.exit(1)
 
-    with open_file_read(args.input_file) as fh:
+    with open_file_read(file_in) as fh:
         yaml_input = fh.read()
 
     if args.prepend_file:
-        if not os.path.exists(args.prepend_file):
+        if not Path(args.prepend_file).exists():
```
kvid, line 325: For some strange reason, this suggestion was not included in my review just now:
(The same argument as @laurierloi describes above also applies here.)
```suggestion
        if not Path(args.prepend_file).is_file():
```

formatc1702:
All comments have been addressed inside #251 or are no longer relevant after 251 is applied

PR #226, opened 2021-03-23
Better additional component BOM table

formatc1702:
Closes #224.
Based on #197 (still unmerged).
Will integrate #214 to allow the user to set `show_part_numbers` and `show_bom_item_numbers` from YAML.

Will rebase this branch onto `dev` once #197 and #214 are merged.

Issue #224, opened 2021-03-20
Improving additional component display, BOM item numbers


formatc1702:
I have been using WireViz at work quite a bit (work being the main reason I haven't had time to contribute to the project 😕) and I have noticed the treatment of "main" components and additional components is not as consistent as it maybe should be.

*Note: The following thoughts might partially contradict previous opinions and decisions... this is what happens when theory meets practice.*

Take the following sample file:

![image](https://user-images.githubusercontent.com/2447251/111868312-8a57f380-8979-11eb-966a-28e0812ff346.png)

<details>
<summary>Source</summary>

```yaml
connectors:
  X1:
    type: Plug
    pincount: 4
    pn: 123
    manufacturer: ACME
    mpn: ABC
    additional_components:
      -
        type: Crimp
        pn: 987
        manufacturer: ACME
        mpn: ZYX
        qty_multiplier: populated
      -
        type: Housing
        pn: 345
        manufacturer: OTHER

  X2:
    type: Receptacle
    pincount: 4
    pn: 234
    manufacturer: ACME
    mpn: DEF
    additional_components:
      -
        type: Crimp
        pn: 876
        manufacturer: ACME
        mpn: WVU


cables:
  W1:
    wirecount: 4
    color_code: DIN

connections:
  -
    - X1: [1-4]
    - W1: [1-4]
    - X2: [1-4]

```

</details>

Issues (easily solvable)

- When using the image file on its own (without the exported BOM), part number info is not visible for additional components. This might be desirable when, e.g. pasting the image file into a larger documentation document. It is probably more consistent to show info for additional components as well.
- Additional components show their assigned item number in the image (`#4` etc.), where as main components (connectors and cables) do not. It should be either both, or none.
  - How about a global option to `show_bom_item_numbers` that then applies to both?
- Personally, I find the `#` symbol visually jarring.. perhaps there is a more elegant solution
  - Maybe [circled numbers](http://xahlee.info/comp/unicode_circled_numbers.html)? ⓪ ① ② ③
    They appear to go at least up to ㊿.
  - Another idea (though potentially cumbersome) would be to use a 1x1 table to make BOM numbers appear in a box.
    This is similar to "balloons" used in mechanical technical drawings
- The additional components section within the component node could be a table with invisible column separators, to get a  more even look when it starts getting more populated.
  - A blank first column for indentation could be a visual cue as to where the list starts and ends.

Other thoughts (discussion required)

- Should user-define BOM item numbers be allowed?
[...]

Issue #183, opened 2020-10-20
[bug] Bundles to have consistent part number fields


Tyler-Ward:
Splitting out of https://github.com/formatc1702/WireViz/pull/115#discussion_r508000264

Curently in a bundle the part number fields will be rendered differently if they are a string or a list. If they are a string they are rendered at the top of a node and if a list they are rendered per wire, it mould make sense for this to be consistent across bundles by always doing the latter.

kvid:
AFAIK, this is by design. Attributes common to all wires are listed above all wires (except image, and notes below), and attributes specified for each wire listed above and below each wire. Currently, both options are implemented for `pn`, `manufacturer`, and `mpn`, but it has been discussed to support the same for `gauge` too.

formatc1702:
> AFAIK, this is by design.

Correct. Here's the history:
- At the beginning, the only parameter that could change on a per-wire basis, was color. Everything else (type, length, gauge, ...) was global for the entire cable, thus displayed above the wires.
- Currently, bundles implement separate [manufacturer] part number parameters.
- In the future (see #56), more parameters might be editable on a per-wire basis.

That's why, IMHO, we shouldn't see this as a bug, nor worry about it too much, because I think some major refactoring may be in order to cleanly implement the transition for more granularity.

Issue #135, opened 2020-07-29
[feature] improved multilingual color input


formatc1702:
## Example

The following YAML input:
```yaml
cables:
  W1:
    gauge: 0.25 mm2
    length: 0.2
    colors: [tomato, ge, 00FF00, BU]
```
should be valid, and produce the following output:

![graph](https://user-images.githubusercontent.com/2447251/88762768-18c10000-d172-11ea-92b3-1fdb9a07dc4b.png)

Yes, tomato should be an accepted color because it's part of the official [HTML color names](https://www.w3schools.com/colors/colors_names.asp) 🍅 

## Motivation

Currently, `wv_colors.py` includes list of german color codes (e.g. `ge` for gelb=yellow), as well as full english color names, but there is no direct way to use them. It would be nice if German users could directly use the german color codes; the directt HTML color and hex options give unlimited flexibility to anybody who needs it.

My proposal is to allow the following inputs:
- English color codes (RD)
- German color codes (rt)
- HTML color entities (red)
- Hex colors (FF0000)

I've chosen to exclude the `#` in the hex colors because YAML treats it as a comment delimiter, and `\#FF0000` is less readable.

## Implementation

The implementation should be quite straightforward.
In addition to a cable's `colors` list, a second `htmlcolors` list (or similar) is needed. Then, a simple parsing function parses the user's input, checking whether it's included in one of the color codes, and setting the appropriate `htmlcolor`, i.e. mapping `RD` to `#ff0000` and `ge` to `#ffff00`, adding the missing `#` if a hex color was specified, and simply passing the input if it is neither of these (`tomato`). If the input is a completely unknown color, it will be rendered black (default GraphViz behavior) but the label will still show the input.

This means that the color parsing would happen in the cable's `__post_init__()` function, and all the GraphViz code generation would no longer use the `translate_color()` function, but read straight from the `htmlcolors` list.

kvid:
These ideas you present, are great! 👍 I just have a few questions about details.

> German color codes (rt)

By accepting German codes as input, I assume requests for supporting additional languages will arrive. As more languages can be supported, it might be hard to avoid ambiguity unless adding a language prefix (e.g. `de.gnge`) or a common `Harness.input_language`.

[...]

PR #251, opened 2021-10-20
Large scale refactoring

17o2:
This is a project wide refactoring effort.
I have decided to address it as one large chunk because the fundamental structure of the code needs to be rebuilt at once for this to work.

Goals:
- Shorter, more atomic functions
- Shallower indentation
- More object-oriented (HTML tags, pins, wires, mates, colors, ...)
- Less hand-crafted HTML, fewer hacky tricks
- Less convoluted code
- Less code repetition

Issues being addressed/solved:
- [x] Implement the ideas initially tried in #249
- [x] #183
- [x] #135
- [ ] #224 
- [ ] #226

A better description will follow once the code is presentable.
I am sharing the work in progress for transparency's sake.


formatc1702:
I have resumed work on the refactoring after a longer period of inactivity. and I would like to propose a course of action to get the bulk of this code integrated into `dev`, and return to incremental improvements through smaller PRs.

- @kvid if you have time, please check that there are no glaring errors when building the example/tutorial files, regarding the graphical output and the BOM.
- Even if the examples build without errors, the code will probably be rough around the edges. Nevertheless, I would then like to merge this PR into `dev` to make it "visible" to the larger group of contributors. And, importantly, this will take a big mental load off of me to know we are back to working in manageable chunks 😺 
- I realize that there are some open discussions and suggestions in the PRs this one is built on top of; i.e. the ones with the [`in latest`](https://github.com/formatc1702/WireViz/pulls?q=is%3Apr+label%3A%22in+latest%22+) tag. I do not want these discussions/suggestions in these PRs to be lost, therefore the plan is:
  - Leave those PRs open for now if they include important unresolved comments.
  - Address the suggestions as new PRs on top of the freshly merged `dev` branch, to avoid nasty rebasing troubles.
  -  Close the old PRs once I feel that everything is addressed.
  This concerns #239, #246,  and to a lesser extent #186, #189. There are no comments on #244, #248, #250; any improvements on these topics should come as new issues/PRs.
- The groundwork has been laid for some other open feature request (such as BOM stuff from #224 + #226).. full implementation will follow in new PRs.

Thank you for your patience and understanding.

kvid:
@formatc1702 wrote:
> I have resumed work on the refactoring after a longer period of inactivity. and I would like to propose a course of action to get the bulk of this code integrated into `dev`, and return to incremental improvements through smaller PRs.
> 
> * @kvid if you have time, please check that there are no glaring errors when building the example/tutorial files, regarding the graphical output and the BOM.

I'd be happy to do that, but the installation failes as described in #287.

> * Even if the examples build without errors, the code will probably be rough around the edges. Nevertheless, I would then like to merge this PR into `dev` to make it "visible" to the larger group of contributors. And, importantly, this will take a big mental load off of me to know we are back to working in manageable chunks 😺

I agree this sounds like a wise decision.

> * I realize that there are some open discussions and suggestions in the PRs this one is built on top of; i.e. the ones with the [`in latest`](https://github.com/formatc1702/WireViz/pulls?q=is%3Apr+label%3A%22in+latest%22+) tag. I do not want these discussions/suggestions in these PRs to be lost, therefore the plan is:
>   
>   * Leave those PRs open for now if they include important unresolved comments.
>   * Address the suggestions as new commits on top of this PR, to avoid nasty rebasing troubles.
>   * Close the old PRs once I feel that everything is addressed.
>     This concerns [Improve output of diagram + BOM as technical drawing #239](https://github.com/formatc1702/WireViz/pull/239), [Replace `os.path` with `pathlib.Path` #246](https://github.com/formatc1702/WireViz/pull/246),  and to a lesser extent [Implement mates between connectors, better autogeneration #186](https://github.com/formatc1702/WireViz/pull/186), [Embed images in SVG output #189](https://github.com/formatc1702/WireViz/pull/189). There are no comments on [Clean up and improve CLI implementation #244](https://github.com/formatc1702/WireViz/pull/244), [Apply `isort` and `black` to source code #248](https://github.com/formatc1702/WireViz/pull/248), [Improve `wireviz.parse()` #250](https://github.com/formatc1702/WireViz/pull/250); any improvements on these topics should come as new issues/PRs.

OK

formatc1702:
One thing I still want to add is updated documentation / changelog. And I might rearrange some code blocks to make more sense.. but no further actual development.

kvid:
I've not yet had time to test all new code, but I list here a few issues I've noticed so far:
- The code building examples still use the `.bom.tsv` extension. I suggest we use a common tuple of supported extensions that also can be available for applications using WireViz as a library.
- I don't like that I now must run `wv_cli.py` from the command line instead of the much more intuitive `wireviz.py`.
- It seems the GV HTML are generated differently now, and I can see some more attributes are generated. We should perhaps create a test case with `bgcolor` declared in all possible ways to verify that the intended logic with default values inherited from the more general level still holds, etc.

formatc1702:
> * I don't like that I now must run `wv_cli.py` from the command line instead of the much more intuitive `wireviz.py`.

The idea is that `wireviz.py` contains the core parsing and output functions (both file and return type), and it's what you would import when using WireViz as a module.
`wv_cli.py` splits off the CLI handling. In practice, users would not call `python3 wv_cli.py`, but simply invoke `wireviz` directly using the [entry point](https://github.com/formatc1702/WireViz/blob/refactor/big-refactor/setup.py#L32). This is also how the usage is described now in the [Readme](https://github.com/formatc1702/WireViz/blob/master/docs/README.md#how-to-run).

I agree that testing `bgcolor` etc. is a good idea, since it's not in the current examples and I haven't used it much myself.

amotl:
> The idea is that `wireviz.py` contains the core parsing and output functions (both file and return type), and it's what you would import when using WireViz as a module. On the other hand, `wv_cli.py` splits off the CLI handling, and gets [materialized as `wireviz` program](https://github.com/formatc1702/WireViz/blob/refactor/big-refactor/setup.py#L32-L36) using a [console script entry point](https://setuptools.pypa.io/en/latest/userguide/entry_point.html).

I like it!

kvid:
It's a long time since I read your very helpful message @amotl, but I forgot to answer properly at the time.

> > The idea is that `wireviz.py` contains the core parsing and output functions (both file and return type), and it's what you would import when using WireViz as a module. On the other hand, `wv_cli.py` splits off the CLI handling, and [gets materialized as `wireviz` program](https://github.com/formatc1702/WireViz/blob/refactor/big-refactor/setup.py#L32-L36) using a [console script entry point](https://setuptools.pypa.io/en/latest/userguide/entry_point.html), when installing the package with `pip`.
> 
> I like it!

I see your point, but it didn't work as intended in my workspace.

> 
> @kvid: I don't know if it's in the documentation yet, but the recommended way to "install" that entrypoint, while in development/sandbox mode, is to invoke `pip install --editable=.` in the repository root directory, where your `setup.py`, `setup.cfg`, or `pyproject.toml` resides.

Thank you for this explanation. That helped me getting a proper development installation and enabled me executing "wireviz" directly instead of using "python wireviz.py" (which stopped working when the entry point changed).

> 
> This will run the installation process in development mode (what was `python setup.py develop` up until recently) and will link the defined entrypoints with your local working tree, so you can run `wireviz` and invoke the code you are currently editing. It is always recommended to do all of that within a Python virtualenv, but I am sure I am telling no news here.
> 
> Edit: I see that [1] may be a bit thin. Let me replicate a full sandbox installation example here if you don't mind:
> 
> ```
> git clone https://github.com/formatc1702/WireViz --branch=refactor/big-refactor
> cd WireViz
> python3 -m venv .venv
> source .venv/bin/activate
> pip install --editable=.
> wireviz --version
> ```

I totaly agree on using Python virtualenv, and that we should recommend it to others. Why not creating a PR where you update this in the [README.md](https://github.com/formatc1702/WireViz/blob/master/docs/README.md)? However, there is a small discrepancy between using `.venv` as folder name and the `venv/` entry in [.gitignore](https://github.com/formatc1702/WireViz/blob/master/.gitignore) - is there a [recommended folder name](https://discuss.python.org/t/trying-to-come-up-with-a-default-directory-name-for-virtual-environments/3750)? My oldest Wireviz workspace uses the more unusual `wv-env` as the folder name (recommended in #43).

On the other hand, maybe it's better to include (several) frequently used folder name alternatives in .gitignore to help the majority of users - even if they use a folder name different from the installation example? What do you think?

nyq:
I am observing two issues with [refactor/big-refactor](https://github.com/formatc1702/WireViz/tree/refactor/big-refactor) branch:

1. `templates` directory is not delivered during installation from `WireViz-refactor-big-refactor/src/wireviz` to [site-packages/wireviz](url). As a result, attempting to process, let's say demo02.yml results in exception because `din-6771.html` could not be found. Manually copying this directory from the source zip file into site-packages/wireviz resolves the problem. `pip3 install WireViz-refactor-big-refactor.zip` command is used to install wireviz on top of fresh Python 3.11.2 installation.
2. When running wireviz with -fP option switch to generate PDF, _PDF output is not yet implemented_ message is displayed. Of course this is not a bug, but it defeats the purpose for me trying this branch as I came here from main branch via #304 to #239 and finally to #251 in search for PDF output.

tests/bom/bomqty.yml
```diff
@@ -0,0 +1,65 @@
...
+        type: Three centimeters
+        qty: 3 cm
+      -
+        type: Takes wirecount times two
+        qty: 2
+        qty_multiplier: wirecount
+      -
+        type: Takes length times three
+        qty: 3 # adding unit here should cause error because the length already has a unit
+        qty_multiplier: length
```
kvid, line 46: The comment "adding unit here should cause error because the length already has a unit" made me wonder how to cope with a use case where I would like to add 5 cable clips per meter as `additional_components` to the cable. Is that possible?

formatc1702: I had not considered the use case of "items per unit of cable length" as opposed to "length as multiple of cable length". I suggest opening a separate issue with this feature request. Ideally, solid unit handling should be implemented first. Then one could even specify cable length in one unit (e.g. km), and the multiplier in another (e.g. 10 per m), with the unit module handling conversion and cancellation of units...

kvid: Such a use case was already described in https://github.com/wireviz/WireViz/pull/115#discussion_r495474219 - and was intentionally enabled then. It is still currently supported in the latest release v0.3.2:
```yaml
connectors:
  X1: &x1
    pincount: 3
  X2: *x1
cables:
  W1:
    wirecount: 3
    colors: [BU,BN,YEGN]
    length: 3 m
    additional_components:
      - type: Clips every 20 cm
        qty: 5
        qty_multiplier: length
connections:
  -
    - X1: [1-3]
    - W1: [1-3]
    - X2: [1-3]
```
cable_clips_html

amotl:
Hi again,

when analyzing the state of this large patch, and the most recent comments, it looks like integrating it might not be too far away. This is my summary what may need to be done here, please extend this list if you think it is missing some other important items.

1. A request from @kvid to improve the developer sandbox documentation at https://github.com/wireviz/WireViz/pull/251#issuecomment-1406714379.
2. Two reports about a packaging flaw and the PDF output feature by @nyq at https://github.com/wireviz/WireViz/pull/251#issuecomment-1483858699.
3. A few details from a self-review by @kvid reported at https://github.com/wireviz/WireViz/pull/251#pullrequestreview-1584509381.
4. Branch shows conflicts and needs to be rebased.

In general, I think it would be totally acceptable to finally merge it, and break out the remaining items into issues, addressing them afterwards one by one. In this case, only item 4. would need to be addressed.

With kind regards,
Andreas.

kvid:
@amotl wrote:
> Hi again,
> 
> when analyzing the state of the onion of this large patch a bit, and looking at the most recent comments, it looks like integrating it might not be too far away? This is my summary what may need to be done here, please extend this list if you think it is missing some other important items to consider.

Is your analysis based only on code and comment inspection, or have you tested the code also. Testing example inputs and comparing the output against output from earlier versions is important, I think.

> 
> 1. A request from @kvid to improve the developer sandbox documentation at [Large scale refactoring #251 (comment)](https://github.com/wireviz/WireViz/pull/251#issuecomment-1406714379).

In my comment you link to, I do ask for this improvement as a separate PR.

> 2. Two reports about a packaging flaw and the PDF output feature by @nyq at [Large scale refactoring #251 (comment)](https://github.com/wireviz/WireViz/pull/251#issuecomment-1483858699).

I would prefer fixing the bug in point 1 of this comment before merging, but I agree point 2 can be addressed as a separate PR.

> 3. A few details from a self-review by @kvid reported at [Large scale refactoring #251 (review)](https://github.com/wireviz/WireViz/pull/251#pullrequestreview-1584509381).

The solution described by @formatc1702 is clearly out of scope or this PR, but maybe a simpler fix can be done. I can take a look when the other things are resolved. However, it's not critical to support such a use case now if that will delay merging significantly.

> 4. Branch shows conflicts and needs to be rebased.

I had to add some bug fixes to `dev` to be able to run some tests using that branch (for comparison), but I'm in the process rebasing #251 on top of them before running new tests of #251.

> 
> In general, I think it would be totally acceptable to finally merge it, and break out the remaining items into issues, addressing them afterwards one by one. In this case, only item 4. would need to be addressed before merging.
> 
> With kind regards, Andreas.

I need more time to run some more tests. I have already detected a few minor things that needs fixing, but I've not yet described them in a review.

kvid:
I pushed yesterday a couple of commits I'm using to:
- Correct `.tsv` filename references to match the rename earlier in this PR.
- Include extra YAML input files when building examples - to detect errors and for verifying the code after rebasing.

formatc1702:
I think the biggest issues that should probably remain inside the scope of this PR (based on a project-wide search for `TODO`) are:

- Part number strings in `gv_additional_component_table()`, `create_graph()`
- Handling additional components in `_add_to_internal_bom()`(see `TODO` there, I can't recall why bundles require a special treatment)
- BOM hash for bundles in `bom_hash()`
- The above-mentioned [packaging flaw](https://github.com/wireviz/WireViz/pull/251#issuecomment-1483858699)

Overall I agree with @amotl 's and @kvid 's assessments. All other `TODO`s can become separate PRs

src/wireviz/wv_dataclasses.py
```diff
@@ -0,0 +1,815 @@
...
+    # their contents are loaded into partnumbers during the child class __post_init__()
+    pn: str = None
+    manufacturer: str = None
+    mpn: str = None
+    supplier: str = None
+    spn: str = None
+    # BOM info
+    qty: NumberAndUnit = NumberAndUnit(1, None)
+    amount: Optional[NumberAndUnit] = None
+    sum_amounts_in_bom: bool = True
```
kvid, line 189: It's not clear to me how to use `amount` and `sum_amounts_in_bom`. They are not mentioned in `syntax.md` and not used in any example/tutorial/test harness. I hope @formatc1702 can provide some examples.

src/wireviz/wv_dataclasses.py
```diff
@@ -0,0 +1,815 @@
...
+    ignore_in_bom: bool = False
+    bom_id: Optional[str] = None  # to be filled after harness is built
+
+    def fill_partnumbers(self):
```
kvid, line 193: Why is this function not called `__post_init__(self)`?

src/wireviz/wv_dataclasses.py
```diff
@@ -0,0 +1,815 @@
...
+        partnos = [self.pn, self.manufacturer, self.mpn, self.supplier, self.spn]
+        partnos = [remove_links(entry) for entry in partnos]
+        partnos = tuple(partnos)
+        self.partnumbers = PartNumberInfo(*partnos)
+
+    def parse_number_and_unit(
+        self,
+        inp: Optional[Union[NumberAndUnit, float, int, str]],
+        default_unit: Optional[str] = None,
+    ) -> Optional[NumberAndUnit]:
```
kvid, line 203: Why is this a method in `class Component` and not a global function? The `self` parameter is not used.

src/wireviz/wv_dataclasses.py
```diff
@@ -0,0 +1,815 @@
...
+        if inp is None:
+            return None
+        elif isinstance(inp, NumberAndUnit):
+            return inp
+        elif isinstance(inp, float) or isinstance(inp, int):
+            return NumberAndUnit(float(inp), default_unit)
```
kvid, line 209: Why is `inp` forced to a `float` here and also in the `str` case below? Why not keep it `int` when possible to avoid the `1.0 x Item` in the output that has been the more natural `1 x Item` before this PR?

src/wireviz/wv_dataclasses.py
```diff
@@ -0,0 +1,815 @@
...
+
+
+@dataclass
+class AdditionalComponent(Component):
+    qty_multiplier: Union[QtyMultiplierConnector, QtyMultiplierCable, int] = 1
+    _qty_multiplier_computed: Union[int, float] = 1
+    designators: Optional[str] = None  # used for components definedi in the
+    #                                    additional_bom_items section within another component
+    bgcolor: SingleColor = None  #       ^ same here
+    note: str = None
```
kvid, line 271: Why use the same class for both `additional_bom_items` and `additional_components`? None of the members in this class seems to be common for these two usages (`designators` only in the former usage, and the rest only in the latter usage)?

What are the disadvantages with defining `class AdditionalBomItems(Component)` and class AdditionalComponent(GraphicalComponent) as separate classes?

src/wireviz/wv_dataclasses.py
```diff
@@ -0,0 +1,815 @@
...
+class TopLevelGraphicalComponent(GraphicalComponent):  # abstract class
+    # component properties
+    designator: Designator = None
+    color: Optional[MultiColor] = None
+    image: Optional[Image] = None
+    additional_parameters: Optional[Dict] = None
+    additional_components: List[AdditionalComponent] = field(default_factory=list)
+    notes: Optional[MultilineHypertext] = None
+    # BOM options
+    add_up_in_bom: Optional[bool] = None
```
kvid, line 321: Why defining the `add_up_in_bom` member? It seems not to be used at all.

src/wireviz/wv_dataclasses.py
```diff
@@ -0,0 +1,815 @@
...
+    # rendering options
+    bgcolor_title: Optional[SingleColor] = None
+    show_name: Optional[bool] = None
+
+
+@dataclass
+class Connector(TopLevelGraphicalComponent):
+    # connector-specific properties
+    style: Optional[str] = None
+    category: Optional[str] = None
```
kvid, line 331: Why defining the `category` member here when it's already defined in a superclass?

.github/workflows/main.yml
```diff
@@ -22,11 +22,11 @@ jobs:
         python -m pip install --upgrade pip
         pip install .
     - name: Create Examples
-      run: PYTHONPATH=$(pwd)/src:$PYTHONPATH cd src/wireviz/ && python build_examples.py
+      run: PYTHONPATH=$(pwd)/src:$PYTHONPATH && python src/wireviz/tools/build_examples.py
```
kvid, line 25: When removing the `cd` command, why still keep the `&&` in front of the `python` command?

src/wireviz/wv_dataclasses.py
formatc1702: currently it is called in the `__post_init__` routine of the derived classes (`Connector`, `Cable`, `AdditionalComponent`).
Since there is no difference between the three, you are right, it could be moved up to the `Component` superclass instead. 

The only case where this *might* be handled differently (if we wanted) is in a bundle, where the `PartNumberInfo` could be passed on to the individual `WireClass` objects in the `wire_objects` dict. Currently this case is handled by `_get_wire_partnumber()`, which works, but there is a reason I added the `# TODO: possibly make more robust/elegant` note there.

For now, let's follow your suggestion and keep the optimizations for later.

src/wireviz/wv_dataclasses.py
formatc1702: No real reason. OK to move to `vw_utils.py` or wherever you see fit.

src/wireviz/wv_dataclasses.py
formatc1702: True. Can be reverted.

I believe the `try: number = float(number)` in the `elif isinstance(inp, str):` section was a primitive way to see whether, after splitting the string into `number` and `unit`, the `number` part was actually a number. This can be rewritten in a way that does not force the conversion. Even `_ = float(number)` would be enough, but I welcome a more elegant check.

src/wireviz/wv_dataclasses.py
formatc1702: No real reason comes to mind. Feel free to split.

src/wireviz/wv_dataclasses.py
formatc1702: It is not used *yet*, but it stems from a wish to be able to specify how to handle lengths in the BOM.

Example:
A harness contains two identical cables of 1m each, another cable of same type but 5m, and two bundles of four identical wires of 1m each.
1. `add_up_in_bom = True` would show the BOM requiring 7m of cable + 8m of wire
  (that is the current behavior)
1. `add_up_in_bom = False` would show the BOM requiring 2 pcs. of cable (1m) + 1 pcs. of cable (5m) + 8 pcs. of wire (1m)

In a workflow where the person asssembling the harness is expecting to cut the cables themselves, option 1 is fine. If they are expecting to receive all cables/wires cut to length, option 2 is the way to go.

Another typical example is heatshrink: If there is a connector that requires 10mm of heatshrink on each of its 15 pins, the BOM could show either
1. 150mm of heatshrink or
2. 15 pcs. of 10mm heatshrink, which is much more understandable in context.

To avoid confusion while it is not implemented, feel free to comment the line out and add a `TODO` for later and I can write a new issue.

src/wireviz/wv_dataclasses.py
formatc1702: Sorry for the confusion. See my [comment in the `add_up_in_bom ` line](https://github.com/wireviz/WireViz/pull/251#discussion_r1359066031).

src/wireviz/wv_dataclasses.py
formatc1702: True. See my [comment](https://github.com/wireviz/WireViz/pull/251#discussion_r1361674997) above for `sum_amounts_in_bom`. Seems this is a duplicate under a different name, and can go.

src/wireviz/wv_dataclasses.py
formatc1702: True. Currently, it would even make sense to define it only in `Cable` class (since it is only used there) instead of the `Component` superclass!

.github/workflows/main.yml
formatc1702: If it works without, throw it out :)

src/wireviz/wv_dataclasses.py
kvid: > True. Currently, it would even make sense to define it only in `Cable` class (since it is only used there) instead of the `Component` superclass!

I agree, but it seems to be used several places outside the `Cable` class. It might be connected to a conflict between two ways of organizing the code:
 - **Object oriented:** 
   * _Encapsulate_ properties and functions that process them into objects with methods.
   * _Abstraction_ - hiding object details that are not needed from the outside.
   * _Inheritance_ between objects when they need a common set of properties and methods.
   * _Polymorphism_ - enable generic algorithms that work on different objects by implementing the code that differ between objects as methods of each object class, and with common method names to be called by the generic algorithms so they don't need to know what kind of object they process.
 - **Procedural programming:**
   * Each procedure/function contains the steps of their algorithm. This is easy to understand as long as the complexity and amount of different alternative value type dependent code don't grow too large.
   * When a lot of the steps in the procedure differ depending on the value type/class being processed, then the code often gets harder to understand due to the needed amount of code to test for different alternatives within each step.

This PR increases the amount of object oriented code by replacing some repeated properties and code with inheritance between classes, but I'm afraid the amount of polymorphism has been reduced, and as a result we see a greater amount of code that test what kind of object is being processed.

Maybe there's a conflict between grouping code together with the algorithm where used, and grouping code together with the object being processed. I feel we should shift this balance a bit in the latter direction in some parts of our code.

formatc1702: as a non-professional programmer, thanks for the input and explaining the dualism. I did indeed move in the OO way, but I am not married to the idea and always open for the best approach for the situation.

.github/workflows/main.yml
kvid: How do you suggest I verify that this change will work?
```suggestion
      run: PYTHONPATH=$(pwd)/src:$PYTHONPATH python src/wireviz/tools/build_examples.py
```
Will the modified code be executed immediately after adding it as a commit to this PR branch, or does it have to be merged into the master branch to be executed?

formatc1702: good question. Does every branch/PR run the workflows defined within itself? One test could be to remove the workflow entirely, or insert a call to a nonexistent module, and see after committing and pushing the change, the checks stop passing. If so, verifying the effect of removing `&&` should be straightforward. If not, leave it as-is for now and test later, e.g. in a test repo.

src/wireviz/wv_dataclasses.py
formatc1702: Moved `parse_number_and_unit` and `NumberAndUnit` definition to `wv_utils.py`.

.github/workflows/main.yml
formatc1702: > Will the modified code be executed immediately after adding it as a commit to this PR branch, or does it have to be merged into the master branch to be executed?

Tested, yes, the change is immediately applied. Removed `&&`, build still works. Resolved.

src/wireviz/wv_dataclasses.py
formatc1702: For now, removed the duplicate definition inside `Connector`, will rely on `Component` superclass for now. See https://github.com/wireviz/WireViz/issues/225#issuecomment-951164193 for potential additional uses of the `category` attribute for connectors.

src/wireviz/wv_dataclasses.py
formatc1702: Fixed.

tobiasfalk:
I do not want to be to anything but is there a time frame of when this is merged, since this is kind of putting new features at hold😁🤦‍♂️

kvid:
When v0.4.1 is released, then those fixes need to be merged with this PR into `dev`. It's a bit hard to guess how long that'll take.

This PR doesn't block all new features, but most HTML generating code is refactored here, so other major code changes in that part of the code based on the latest release (before this PR) will need rework before a merge-in after this PR.

tobiasfalk:
Well, since it renames "Dataclasses.py" and "Harnesses.py" everything that needs to be implemented in either of those two has to be redone.
So is it a problem if I redo my changes from #369 and #379 based on this branch, and maybe even create a merge to this branch?
Mainly because I currently have time, and I do not know how much time I have when this is merged.
And for the future, when renaming of a file is done, this should be its own merge and this should be processed a bit faster so that other people do not have to reimplement their changes.
Ps. Sorry If I am a bit annoying.

kvid:
If you have time for rework now, but maybe not later, then I suggest you consider creating two new PRs based on the `dev` branch, but with feature branches that you branch off from the current #251 branch. Then you have effectively included all current #251 changes in your new feature branches before reworking your old feature branches on top of each new feature branch. If you also need any of the fixes to be released as v0.4.1, then you can cherry-pick that commit into your new feature branch before committing your code that depends on it.

This way it should be much easier to later rebase your new feature branches on top the new `dev` after the future #251 merge-in, even if the #251 feature branch has been rebased, force-pushed, and merged with v0.4.1 in the meantime. Such a rebase should remove all #251 changes and cherry-picked fixes from your feature branch unless a conflict needs resolving first. I recommend you keep both the old and new PRs as drafts until you prepare for actual merge-ins. Don't close your old PRs until successful merge-ins of the new ones.

Be aware that it's not safe to base your new PRs on the #251 feature branch, because when that branch is deleted after the #251 merge-in, all PRs based on a deleted branch will be automatically closed by github, and it's not easy to reopen such automatically closed PRs.

> Ps. Sorry If I am a bit annoying.

I don't find you annoying! 😄 Maybe you mean "_annoyed_"?

tobiasfalk:
Thx
Is this what you meant? https://github.com/tobiasfalk/WireViz/tree/Jumpers_reimp

kvid:
@tobiasfalk wrote:
> Thx Is this what you meant? https://github.com/tobiasfalk/WireViz/tree/Jumpers_reimp

That seems to be the result of creating `Jumpers_reimp` by branching it off at `dev` with a command like this: `git branch --no-track Jumpers_reimp remotes/{official remote}/dev` and then merged `remotes/{official remote}/refactor/big-refactor` into it.

In that case, I don't know if a later rebase of `Jumpers_reimp` on top of `dev` after #251 has been merged in, will remove the #251 changes. Personally, I usually avoid using merge unless I need to document joining two branches into one. When joining changes from two branches without such a need, I usually prefer cherry-picking and rebasing instead of merging, because the commit history becomes much easier to read, and `git rebase` automatically removes redundant commits.

I don't know the remote names of your local repo, but you can list them with `git remote -v`. Replace `{official remote}` in my commands with your remote name for our official public github repo, and `{personal fork}` with your remote name for your personal fork at github copied from our official public github repo.

These are the commands that represent what I suggested in my previous comment (`new_feature1` can be whatever name you want):
```sh
git branch --no-track new_feature1 remotes/{official remote}/refactor/big-refactor
git push -u {personal fork} new_feature1
```

formatc1702:
> I do not want to be to anything but is there a time frame of when this is merged, since this is kind of putting new features at hold😁🤦‍♂️

I fully agree and I feel bad that it has taken so long; it's a big change that I started on my own, but due to limited capacity I have not been able to complete it in a satisfactory way.

I am extremely thankful for all input recently, and the big help in developing, reviewing and integrating fixes. Currently, I can offer little more than some light guidance and clicking the final "merge" on PRs, or publishing a release to PyPI, but not much coding.

#365  is almost ready, and merging this refeactoring should be top prio afterwards. Again, I will appreciate the support here (especially from contributors more versed in resolving the ugly merge conflicts that may occur) as well as in the subsequent bugfixing and integrating all fixes made to the old codebase in the meantime (from v0.4 to v0.4.1).

So, big ❤️  to the active contributor community.

tobiasfalk:
#365 is merged, I would suggest we now merge this one and create a Task List to make working on this easier.

tobiasfalk:
> If you have the time and the enery, you are welcome to do your own review in parallel, maybe describe what you tested, and suggest alternatives.

I will look in to it, but I do not relay have much experience in code testing and reviewing.

> My idea is to bring in the v0.4.1 changes stepwise into `dev` after this PR because some of the v0.4.1 changes need to be reimplemented due to the refactored code. What do you think?

Sounds good, as mentioned, I would start with a simple List of what needs to be transplanted/reimplemented.

Two of the big points that I would put on this Task List, would be  #406 and #387, since they would make communication with the users easier and take the work off the developers, yes even though they practically have nothing to do with this PR.

tobiasfalk:
GitHub has something called Projects (here the FreeCAD Git for example: https://github.com/orgs/FreeCAD/projects) this would be a good solution to group the relevant Issues and so.

tobiasfalk:
I just ran the examples and looked over it and found that with tutorial one the wires are not drawn correctly:
grafik
this does not happen when a color is defined:
grafik

Can easly be fixed with `color = "#000000:#FFFFFF:#000000"` in line 402 of wv_graphviz.py and 
```
def gv_wire_cell(wire: Union[WireClass, ShieldClass], colspan: int) -> Td:
    if wire.color:
        color_list = ["#000000"] + wire.color.html_padded_list + ["#000000"]
    else:
        color_list = ["#000000","#FFFFFF","#000000"]

    wire_inner_rows = []
    for j, bgcolor in enumerate(color_list[::-1]):
        wire_inner_cell_attribs = {
            "bgcolor": bgcolor if bgcolor != "" and bgcolor != "#000000" else "#000000",
```

starting at line 366

tobiasfalk:
Also, something with the Line Thickness of the wire block is not right:
grafik

grafik

this can also be solved rather simply with seting `tbl = Table(rows, border=1, cellborder=0, cellspacing=0)` at 362 wv_graphviz.py

src/wireviz/wv_graphviz.py
```diff
@@ -0,0 +1,632 @@
...
+
+    return wire_outer_cell
+
+
+def gv_edge_wire(harness, cable, connection) -> Tuple[str, str, str, str, str]:
+    if connection.via.color:
+        # check if it's an actual wire and not a shield
+        color = f"#000000:{connection.via.color.html_padded}:#000000"
+    else:  # it's a shield connection
+        color = "#000000"
```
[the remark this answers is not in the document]
formatc1702: Back in the day, drawing wires as thin when they don't have an explicit color assigned, was intentional, to distinguish an "uncolored" wire from a white or black one. For example, shields are also drawin in this style, since they don't have a color per se.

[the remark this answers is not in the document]
tobiasfalk: OK, didn't know that.

tobiasfalk:
@kvid and @formatc1702, how does it stand with merging this and creating a todo list of things from 0.4.1 that need to be reimplemented?

ssweber:
Has there been any discussion to unblock this? The WIP shorts/jumpers and wire-banding would really make WireViz a great output format to document control cabinets.

formatc1702:
Unfortunately, all WireViz development* is on hiatus from my side due to personal reasons. @kvid is working on this as time allows, and has merging rights. I'm sorry I don't have better news.

\* I can still take care of PyPI publishing.

kvid:
The latest issue I've worked on, is to find the missing steps to fulfill #408 (see my post-merge-in comments there) - that is currently blocking all PRs - including this big one. However, unforeseen circumstances have forced me to put most of my effort on hold for a period. During the last weeks, I've only been able to answer a few issue requests.

17o2:
Closing PR after using #447 to simplify the merge process.

PR #254, opened 2021-10-22
Improve gracefulness when invoking `wireviz.parse()` without `file_out` argument

### LINKED ISSUE #253: WireViz 0.3: `file_out` argument to `wireviz.parse()` became mandatory
Hi Daniel,

as a followup to #252, while working on https://github.com/daq-tools/wireviz-web/pull/3, I would like to bring another detail to your attention.

Specifically, it revolves around the patch https://github.com/daq-tools/wireviz-web/pull/3/commits/e9148b49 we needed to satisfy the new mandatory requirement of `wireviz.parse()` to obtain the `file_out` argument.

As the code is focused on getting a full `Harness` instance back by signalling `return_types="harness"`, it was a bit surprising that this argument became mandatory, given that it is announced as optional in the function signature. Fair enough, no worries about the breakage per se, we are playing 0.x releases here.

However, I would like to ask / propose if it would be worth making this argument optional again? Following the code, I discovered only two places where it is actually used: [1] and [2]. The first occurence uses the filename to deduce a title. The second one forwards it as `gv_dir` to an instance of `Image`. While the second one honors `None`ness through `file_out if file_out else ''`, the first one `Path(file_out).stem` does not.

Instead, it croaks with
```python
  File "/Users/amo/dev/daq-tools/sources/wireviz-web/wireviz_web/core.py", line 119, in wireviz_render
    harness: Harness = wireviz.parse(yaml_input=input_yaml, return_types="harness")
  File "/Users/amo/Library/Caches/pypoetry/virtualenvs/wireviz-web-UCc3rrkF-py3.9/lib/python3.9/site-packages/wireviz/wireviz.py", line 44, in parse
    harness.metadata['title'] = Path(file_out).stem
  File "/usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1082, in __new__
    self = cls._from_parts(args, init=False)
  File "/usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 707, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 691, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

With kind regards,
Andreas.

[1] https://github.com/formatc1702/WireViz/blob/v0.3/src/wireviz/wireviz.py#L44
[2] https://github.com/formatc1702/WireViz/blob/v0.3/src/wireviz/wireviz.py#L57

amotl:
Hi Daniel,

this patch would be a fix for the regression we reported and would resolve #253. Let me know what you think about it.

With kind regards,
Andreas.

### DIFF
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -40,8 +40,13 @@ def parse(yaml_input: str, file_out: (str, Path) = None, return_types: (None, st
         options = Options(**yaml_data.get('options', {})),
         tweak = Tweak(**yaml_data.get('tweak', {})),
     )
+
+    # When title is not given, either deduce it from filename, or use default text.
     if 'title' not in harness.metadata:
-        harness.metadata['title'] = Path(file_out).stem
+        if file_out is None:
+            harness.metadata['title'] = "WireViz diagram and BOM"
+        else:
+            harness.metadata['title'] = Path(file_out).stem
 
     # add items
     sections = ['connectors', 'cables', 'connections']


amotl:
Hi Daniel,

thanks for releasing WireViz 0.3.1. After bringing in https://github.com/daq-tools/wireviz-web/pull/4, I also just published WireViz-Web 0.3.1.

With kind regards,
Andreas.

PR #288, opened 2022-08-09
Loops by pin or pinlabel, loop color

visatu:
Added possibility to make loops based on `pin` or `pinlabel`, and new `loop_color` attribute to connector.

Input
```yaml
connectors:
  A:
    pins: [1, 2, 3, A, B, C]
    pinlabels: [ONE, TWO, THREE, AA, BB, CC]
    loops:
      - [1, 2]          # pin-pin
      - [TWO, THREE]    # label-label
      - [AA, C]         # label-pin
    loop_color: "RD"
```
... results in:
image

Trying to link based on names that have none or more than one results when searching `connector.pins` or `connector.pinlabels` results in an error.

This PR includes the newlines added in #261, output of the dev branch woudln't make any sense without that.


visatu:
About loop 'length' which becomes an issue with bigger connectors with loops, as the loops seem to grow with the size of the diagram (?) and start overlapping with other connecting wires as well as each other.

Tried to play around setting the graphviz `minlen` attribute for the `edge` connections of the loops, to make the loop size smaller of possibly adjustable. I'm not experienced in using graphviz _at all_ so this was basically a case of trial and error.

Tried so far:
 - setting `minlen` in different points of `Harness.py` where `dot.attr` is called, such as [Harness.py:198](https://github.com/visatu/WireViz/blob/adb3ce8fe2ab266650b39f3d0a57635be62c1d6e/src/wireviz/Harness.py#L198)
 - setting `minlen` where `dot.edge` is called such as [Harness.py:213](https://github.com/visatu/WireViz/blob/adb3ce8fe2ab266650b39f3d0a57635be62c1d6e/src/wireviz/Harness.py#L213)
 - Creating subgraphs for loops and setting `minlen` in the edge attributes in there.
 
Depending on where `minlen=0` or `minlen=1` is set, the output is either unchanged or _all_ of the `edge`s in the diagram become minimized:
image

 From this experimentation it seems that the `minlen` is set globally and the last set value is what decides the `minlen` of the whole graph?

kvid:
I like most of these ideas. However, I feel we also need more attributes than color for each loop wire - e.g. wire entries with zero length makes no sense in the BOM. How about specifying the loop wires as a bundle of wires (with normal wire attributes)? Such a bundle could either be specified with a name together with other cables (and refer to this name from the connector) or maybe preferably directly under a `loop_wires` attribute for the connector. The bundle box could then be placed close to the connector box in the diagram (by placing an invisible group around them) or optionally joined with the connector box, but loop wires should still be drawn directly between connector pins.

PR #298, opened 2022-12-12
Add unpopulated option to additional components qty multiplier

JeremyRuhland:
qty_multiplier can be set to `unpopulated` when defining additional connector components. This will calculate the number of empty pin spaces in the connector.

I use this for waterproof connectors which require sealing plugs to be installed in unoccupied sockets.

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -18,7 +18,7 @@ MultilineHypertext = (
 Designator = PlainText  # Case insensitive unique name of connector or cable
 
 # Literal type aliases below are commented to avoid requiring python 3.8
-ConnectorMultiplier = PlainText  # = Literal['pincount', 'populated']
+ConnectorMultiplier = PlainText  # = Literal['pincount', 'populated', 'unpopulated']
 CableMultiplier = (
     PlainText  # = Literal['wirecount', 'terminations', 'length', 'total_length']
 )
@@ -232,6 +232,8 @@ class Connector:
             return self.pincount
         elif qty_multiplier == "populated":
             return sum(self.visible_pins.values())
+        elif qty_multiplier == 'unpopulated':
+            return (self.pincount - sum(self.visible_pins.values()))
         else:
             raise ValueError(
                 f"invalid qty multiplier parameter for connector {qty_multiplier}"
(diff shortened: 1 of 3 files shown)


kvid:
@JeremyRuhland - Thank you for your suggested feature implementation, and I wonder why you closed this PR just a few hours after creating it in December? I find your use case valuable and your implementation fit well together with existing features.

I can see that you changed the base branch from `master` to `dev` and then back again just a few minutes after creating the PR, and I agree that `dev` should be the base branch. Maybe you discovered warnings about conflicts after changing to `dev` and therefore changed back to avoid the conflicts. Such conflicts appear because there are a lot of commits in the `dev` branch that are not yet in the `master` branch, and I recommend rebasing your feature branch on top of the current `dev` branch:
- Any uncommitted changes in your workspace should be committed or stashed away before rebasing.
- The command `git rebase --onto dev master unpop_addit_cmp` will try to replay on top of `dev` all your changes in `unpop_addit_cmp` that is currently on top of `master`. Often, git manages to do most of this automatically, but in your case two conflicts had to be resolved during the rebase process. See [the result in my fork](https://github.com/wireviz/WireViz/compare/dev...kvid:unpop_addit_cmp) and verify that it still contains your intended changes.
- After such a rebasing, the rebased branch must be force-pushed to your fork (`git push -f`) to update the PR.
- Then the base branch must be set to `dev` and there should be no more conflicts unless someone has already added more commits to `dev` since the state of `dev` at the time starting the rebase. In that case the process must be repeated once more, but the number of conflicts is normally less this time.

Unless you rather want to do the whole process yourself, I can normally push my rebase result onto your fork to save you the work I have already done (unless you unchecked "Allow edits and access to secrets by maintainers" when creating your PR). If the weeks go without you responding here, I might just push my result to your fork to be able to make a proper review of your PR.

src/wireviz/DataClasses.py
```diff
@@ -232,6 +232,8 @@ def get_qty_multiplier(self, qty_multiplier: Optional[ConnectorMultiplier]) -> i
             return self.pincount
         elif qty_multiplier == "populated":
             return sum(self.visible_pins.values())
+        elif qty_multiplier == 'unpopulated':
+            return (self.pincount - sum(self.visible_pins.values()))
```
kvid, line 236:
```suggestion
            return max(0, self.pincount - sum(self.visible_pins.values()))
```
As already reasoned in https://github.com/wireviz/WireViz/pull/298#issuecomment-1719853330, I suggest limiting this multiplier value to minimum zero.

formatc1702:
Changelog updated immediately post-merge.

Issue #316, opened 2023-05-15
Project maintainer wanted


formatc1702:
As you may have noticed, I have not been able to work on the project or even engage in discussions here for a while now. Due to shifting interests, limited time, and other factors, I currently cannot maintain WireViz at the level I feel it deserves.

Therefore I am open to talking with anyone interested about handing over the duties as main project maintainer. Obviously my first priority would be somebody out of the existing contributor pool. Anyone interested please reach out here (preferred for transparency), or privately via the e-mail in my user profile.

We would need to figure out how exactly the delegation should happen.
The two alternatives that come to mind are full control over the mainline repo, which would require a very high level of trust, or defining a straightforward process of submitting finished, tested* and documented PRs that I can simply green-light and merge.
(* Ideally tested by another contributor)

My hope is that losing the burden of being the primary maintainer could help me re-engage with the project in a different way. I have ideas for future features and certain wishes for the "look&feel", which I would like to share but do not have the energy and resources to commit to implementing and testing myself.

I look forward to any comments.

As a separate point, I would like to thank @kvid for their active involvement in responding to other users' issues, and past efforts in reviewing PRs. 🙏

kvid:
Thank you, @formatc1702 for describing your situation and suggesting different paths to bring this project further. I've not been as active as I want myself either lately, but I'm willing to help as long as the tasks don't get too big. The number of changes in #251 is a bit overwhelming, and I therefore have not been able to get the oversight that I feel is needed to make a proper review. What is the minimum effort needed to merge this large PR into `dev` to avoid it blocking other PRs, and how can we all join our efforts to fulfill the needed steps?

formatc1702:
I had been testing the refactored code using the existing examples and demos, as well as some other test files, and it appeared to be working fine except for some particular points marked as `TODO` in the code. I would need to play with the code again after all this time to give a particular example.

The changelog needs to be updated. I want to get rid of the `changelog` branch and go back to updating `CHANGELOG.md` within each PR again.

Currently the code history is `dev` -> `latest` -> `big-refactor`, roughly speaking.

I would be comfortable with merging `latest` into `dev` asap, without further review, and updating the changelog accordingly, since this is simply a stack of multiple cleanly separated PRs independent of the refactoring.

[...]

PR #318, opened 2023-06-01
Use portable OS error codes so program doesn't crash

RedshiftVelocities:
I think OSError had an update at some point, and the error code the program looks for changed. Anyhow this change fixes the titleblock-added version of the program.

### DIFF
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -390,7 +390,7 @@ def _get_yaml_data_and_path(inp: Union[str, Path, Dict]) -> (Dict, Path):
             # if no FileNotFoundError exception happens, get file contents
             yaml_str = open_file_read(yaml_path).read()
         except (FileNotFoundError, OSError) as e:
-            # if inp is a long YAML string, Pathlib will raise OSError: [Errno 63]
+            # if inp is a long YAML string, Pathlib will raise OSError: [errno.ENAMETOOLONG]
             # when trying to expand and resolve it as a path.
             # Catch this error, but raise any others
             from errno import ENAMETOOLONG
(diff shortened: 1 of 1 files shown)


kvid (changes requested):
1. I suggest using the [Standard errno system symbols](https://docs.python.org/3/library/errno.html) to avoid numeric constants that differ between platforms.
2. We might need to test for other errors as well. I managed to get this error when trying an 80000 character dummy string in Python 3.9.13 at Windows 10: `ValueError: _getfinalpathname: path too long for Windows`
3. This is a typical problem where it helps a lot to have unit tests for different edge cases to make it easier to execute the same tests at all platforms (often several developers are involved because not all developers have access to all platforms). I suggest we start collecting such test code. Until we get a better structure for test code, I suggest some clearly marked unit test function at the bottom of the source file containing the major part of the code to be tested.

RedshiftVelocities:
For my own use, is there currently any version of the program that supports both the generation of drawings with a titleblock as well as autogeneration of connectors?

kvid:
@RedshiftVelocities I just found out to my surprise that this PR was closed. I believe it was not intentional, but rather an automatic consequence of deleting the base branch of your PR (it's an unfortunate functionality of github).

This base branch was one of several branches that got deleted when merging their PRs into the `dev` branch (all PRs from the `latest` branch - see also #316 for a description of the process). I expect the current `dev` branch to contain what you need for your own use, but I've not yet been able to test it myself. PR #251 is not yet merged into the dev branch, so your PR is not obsolete. 

It might be hard to reopen such an automatically closed PR, but you should be able to rebase your branch on top of the current `dev` branch, then create a new PR with `dev` as base branch, and call it a continuation of this PR.

kvid:
I managed to re-create the deleted base branch at the exact same commit as it was when deleted earlier this summer. That enabled me to re-open this PR. Then I changed the base branch to `dev` because all commits from the old base branch has already been merged into `dev`. More commits have later been added on top of `dev`, so I rebased your PR branch on top of the current `dev` to prepare a merge. I'm sorry if force-pushing your branch causes any trouble for your local branch, but I have a backup of the previous state, so It will be possible to undo it if needed.

kvid (approved):

PR #339, opened 2024-05-05
Prepare v0.4 release

formatc1702:
[As proposed](https://github.com/wireviz/WireViz/issues/101#issuecomment-2077231783), let's release v0.4 before integrating the large PR #251 into the `dev` branch.

This PR is intended to do a final check before release, to address any inconsistencies.
The obvious one would be to add the release date to the changelog.


src/wireviz/wv_colors.py
```diff
@@ -3,181 +3,198 @@
...
+    "TEL": [  # 25x2: Ring and then tip of each pair
+        "BUWH", "WHBU", "OGWH", "WHOG", "GNWH", "WHGN", "BNWH", "WHBN", "SLWH", "WHSL",
+        "BURD", "RDBU", "OGRD", "RDOG", "GNRD", "RDGN", "BNRD", "RDBN", "SLRD", "RDSL",
+        "BUBK", "BKBU", "OGBK", "BKOG", "GNBK", "BKGN", "BNBK", "BKBN", "SLBK", "BKSL",
+        "BUYE", "YEBU", "OGYE", "YEOG", "GNYE", "YEGN", "BNYE", "YEBN", "SLYE", "YESL",
+        "BUVT", "VTBU", "OGVT", "VTOG", "GNVT", "VTGN", "BNVT", "VTBN", "SLVT", "VTSL",
+    ],
+    "TELALT": [  # 25x2: Tip and then ring of each pair
+        "WHBU", "BU", "WHOG", "OG", "WHGN", "GN", "WHBN", "BN", "WHSL", "SL",
+        "RDBU", "BURD", "RDOG", "OGRD", "RDGN", "GNRD", "RDBN", "BNRD", "RDSL", "SLRD",
```
kvid, line 32: I suggest keeping the original extra spaces to align the single color codes with the double-color codes below. This was probably lost in a Black formatting, but as such a formatting is now turned off, we should keep the spaces.
```suggestion
        "WHBU", "BU",   "WHOG", "OG",   "WHGN", "GN",   "WHBN", "BN",   "WHSL", "SL",
        "RDBU", "BURD", "RDOG", "OGRD", "RDGN", "GNRD", "RDBN", "BNRD", "RDSL", "SLRD",
```
I have already written this comment earlier, but now it seems to be lost. I wonder if you'll get them both?

kvid (changes requested):
See my suggested changes after looking through the major changes . One of them seemed to be lost, so I had to write it again. I hope I didn't loose any other comments.

kvid:
See my bottom question in https://github.com/wireviz/WireViz/pull/309#issuecomment-2095708116 where I point out that the code to be released now only lists quite old Python versions in `setup.py` and our github workflow. However, I fully understand that you want to avoid including new Python versions in the last minute unless you already have experience using this codebase in that Python version and are confident it has been working for a while with different use cases.

src/wireviz/wv_colors.py
kvid: Now, I see both my comments about the same suggestion, and I wonder why I didn't when submitting the review. If you also see both, then you can ignore this latter copy. 😀

kvid:
After thinking a bit, I now believe it's best to leave the supported Python versions unchanged for this release. #251 has already such changes, and we'll get more time testing it after that is merged into `dev`. The workflow already test 3.8, and all later minor versions should normally be backwards compatible. I think we therefore can expect them to work with this release as well.

src/wireviz/wv_cli.py
```diff
@@ -0,0 +1,149 @@
...
+    "s": "svg",
+    "t": "tsv",
+}
+
+epilog = "The -f or --format option accepts a string containing one or more of the "
+epilog += "following characters to specify which file types to output:\n"
+epilog += ", ".join([f"{key} ({value.upper()})" for key, value in format_codes.items()])
+
+
+@click.command(epilog=epilog, no_args_is_help=True)
```
kvid, line 31: Is there a good reason for removing the `-h` option for help?

If not, then I suggest adding it in addition to `--help`. Then we have both help options unchanged from our previous release, and it's consistent with our other new options that all have both short and long options.
```suggestion
@click.command(
    epilog=epilog,
    no_args_is_help=True,
    context_settings=dict(help_option_names=["-h", "--help"]),
)
```

docs/CHANGELOG.md
```diff
@@ -1,17 +1,43 @@
 # Change Log
 
+## [0.4](https://github.com/formatc1702/WireViz/tree/v0.4) (2024-XX-XX)
+
+### Backward-incompatible changes
+- New syntax for autogenerated components ([#184](https://github.com/wireviz/WireViz/issues/184), [#186](https://github.com/wireviz/WireViz/pull/186))
+  - Components that are not referenced in any connection set will not be rendered. Instead, a warning will be output in the console. ([#328](https://github.com/wireviz/WireViz/issues/328), [#332](https://github.com/wireviz/WireViz/pull/332))
+- New command line interface ([#244](https://github.com/wireviz/WireViz/pull/244)). Run `wireviz --help` for details
```
kvid, line 8: You might want to mention both `-h` and `--help` alternatives (or only mention `-h`), but it might be equally good to leave it unchanged (maybe the long option is easier to remember for the inexperienced user).

docs/README.md
```diff
@@ -135,17 +137,16 @@ mywire.bom.tsv    BOM (bill of materials) as tab-separated text file
...
+```
+$ wireviz ~/path/to/files/*.yml
+```
 
-### Syntax description
+To see how to specify the output formats, as well as additional options, run:
 
-A description of the WireViz YAML input syntax can be found [here](syntax.md).
+```
+$ wireviz --help
```
kvid, line 148: You might want to mention both `-h` and `--help` alternatives (or only mention `-h`), but it might be equally good to leave it unchanged (maybe the long option is easier to remember for the inexperienced user).

kvid (changes requested):
I assume removing the `-h` CLI option is not intentional.

docs/CHANGELOG.md
```diff
@@ -24,19 +50,21 @@
 - Add harness metadata section ([#158](https://github.com/formatc1702/WireViz/issues/158), [#214](https://github.com/formatc1702/WireViz/pull/214))
 - Add support for supplier and supplier part number information ([#240](https://github.com/formatc1702/WireViz/issues/240), [#241](https://github.com/formatc1702/WireViz/pull/241/))
 - Add graph rendering options (colors, font, color name display style, ...) ([#158](https://github.com/formatc1702/WireViz/issues/158), [#214](https://github.com/formatc1702/WireViz/pull/214))
```
kvid, line 52: These two bullet points seem to be duplicated just below. Normally, I would suggest deleting the latter copy, but I've already suggested a minor change to the latter copy in my first review 3 days ago, so I suggest deleting this copy instead:
```suggestion
```

src/wireviz/DataClasses.py
```diff
@@ -154,52 +174,74 @@ def __post_init__(self) -> None:
...
 
         for loop in self.loops:
-            # TODO: check that pins to connect actually exist
             # TODO: allow using pin labels in addition to pin numbers, just like when defining regular connections
             # TODO: include properties of wire used to create the loop
             if len(loop) != 2:
-                raise Exception('Loops must be between exactly two pins!')
+                raise Exception("Loops must be between exactly two pins!")
+            for pin in loop:
+                if pin not in self.pins:
```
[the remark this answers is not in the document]
kvid: PR #288 claims to include such functionality, but it seems a bit unfinished, and I haven't checked how easy it might be to cherry-pick only that functionality. Unless you know about a quick-fix with existing code, then I guess we should postpone this functionality.

kvid (changes requested):
I suggest expanding the "Installing the development version" section of README to include recommending `venv` by, e.g. copying commands from https://github.com/wireviz/WireViz/pull/251#issuecomment-1278163024 (checking out the `dev` branch) and also include `pip3 install isort` and `pip3 install black` after activating `.venv`. It will lower the threshold for potential contributors to have a list of commands to get started.
I don't know if it makes sense to create a `requirements-dev.txt` like this to simplify the README description:
```
-r requirements.txt
isort
black
```
Edit: This task about installation description can of course wait unless you have a sequence of commands ready that you have used yourself and don't need testing.

docs/CHANGELOG.md
formatc1702: Let's leave it as is here.

docs/README.md
formatc1702: Let's leave as is for simplicity.

src/wireviz/DataClasses.py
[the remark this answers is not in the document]
formatc1702: Let's postpone please.

formatc1702:
Thank you for your review, I have integrated most suggestions and written a comment on the ones that I didn't.

> After thinking a bit, I now believe it's best to leave the supported Python versions unchanged for this release.

Agreed.

> I suggest expanding the "Installing the development version" section of README [...] It will lower the threshold for potential contributors to have a list of commands to get started.

Good idea, but let's do this after the refactor is merged into `dev`, to avoid PRs based on the v0.4 master branch i.e. on the non-refactored code. A PR to update the README, using the refactored branch as a base, is no problem.

Once v0.4 is merged into master and published, I will try to cherry-pick all of your code suggestions (as far as they are still applicable) to the refactored code so we don't suffer a regression there.

Since most of the current suggestions are quite minor, would you give the OK for releasing v0.4? Then all further review/suggestion effort can be focused on the newer code. Thanks!

kvid (dismissed):

kvid:
Thank's for accepting most of my suggestions.

Note that I pushed a small commit that you should approve or undo. I think it'll help users with old YAML files to get such a notice instead of something that looks like an internal error.

formatc1702:
It's a good idea. Let's see if this can be maintained/updated for future changes; for now it should at least catch the big issues, which is already helpful!

formatc1702:
It's a good idea. Let's see if this can be maintained/updated for future changes; for now it should at least catch the big issues, which is already helpful!

formatc1702:
I am traveling and will be able to merge/publish next week.

kvid (commented):
I just discovered two of the new examples had "0 m" as cable quantity in the BOM because of missing cable `length` attributes. If this is an unintentional error, it's very easy to correct before the release, but It's no big deal to me if you intentionally leave it as it is.

Maybe a waning about each zero cable/bundle length in the harness could be a future feature?

formatc1702:
Let's leave the examples as they are.
For the future, not specifying any length could indeed trigger a warning, and at the same time, the BOM entry should not list 0m, but simply stay blank.

formatc1702:
v0.4 is released 🎉  The next step is somehow integrating #251 into `dev`...

kvid:
@formatc1702 wrote:
> v0.4 is released 🎉 

Congratulations! Thank's a lot for your effort.

> The next step is somehow integrating #251 into `dev`...

First, we need to include all the latest commits from the release into `dev`. Do you prefer merging `master` into `dev` or rebasing `dev` on top of `master`? In the latter alternative, do we risk auto-closing other PRs due to force-pushing their base branch?

After that step, we should rebase the #251 branch on top of the updated `dev`.

formatc1702:
Another alternative would be to cherry pick some changes that went into master but not into dev (such as your latest suggestions).

I don't think a rebase/merge makes sense since a lot of the changes are not 1:1 applicable to the refactored code anyway. This way, we avoid weird cross-merges as well.

Issue #325, opened 2023-09-20
[feature] Add grouping for connectors to represent f.e. PCB with multiple connectors


Heenne:
Currently I am documenting the cable harness for my 3D-printer. In this printer is a controller board with over 20 connectors. If I add all these connectors they are not very well grouped to visualize that these connectors are all on one single PCB. Therefore, I would like to propose the following feature:

Add possibility to create a box with a name around a list of connectors to represent and better visualize a single component with multiple connectors.

Thank you for this awesome tool, it is real fun to work with it!

kvid:
Thank you for sharing your use case. It is possible to group several connectors and/or cables and draw a frame around them with a title text by adding an entry in the `tweak` section. See an example in https://github.com/wireviz/WireViz/issues/276#issuecomment-1041898460 where one connector and one cable is grouped, but you can include many of each in the same group. Notice that connector/cable names need to be quoted when they contain certain characters, e.g space.

Heenne:
Thank you very much for the quick response. The example you linked is exactly how I imagined it.
If I got time, can I make a pull request to add a simple example for this to the other examples?

JarrettR:
Also requesting this.

The linked example is a good workaround, but not quite production ready.
The main issues are that it only groups connectors vertically, and (IMO) `tweak` should be avoided except in the edgiest edge cases.


Here's an example, trimmed down from a WIP of mine:
```

templates:
  - &molex_fan
    type: Molex KK 254
    subtype: female
    pinlabels: [GND, +12V, TACH, FAN]
    show_pincount: false
  - &ps_small
    type: Molex KK 254
    subtype: female
    pincount: 6
  - &ps_large
    type: Molex KK 254
    subtype: female
    pincount: 22
connectors:
  JB-J22:
    <<: *molex_fan
    notes: "Fan 1"
  JB-J23:
    <<: *molex_fan
    notes: "Fan 2"
  JB-J2:
    <<: *molex_fan
    notes: "Fan 3"
  JB-J24:
    <<: *molex_fan
    notes: "Fan 4"
  JB-J25:
    <<: *molex_fan
    notes: "Fan 5"
  JB-J18:
    <<: *ps_small
    notes: "PS L 1"
  PS-L-J18:
    <<: *ps_small
  JB-J19:
    <<: *ps_large
    notes: "PS L 2"
  PS-L-J19:
    <<: *ps_large
  Fan:
    type: 'Fan'
    pincount: 4
  PS-L-1:
[...]

PR #357, opened 2024-05-22
Add optional tweak per node with name placeholder

### LINKED ISSUE #349: [feature] Add tweak option per node
I suggest adding a tweak option for each node (both cables and connectors). It has been on mind many times, but was now triggered again by @martinrieder, and I create a new issue to avoid cluttering the original issue where such a feature would be helpful.

@martinrieder wrote:
> [...]
> My suggestion on this would take some universal approach: allow the tweak to be defined beneath any cable or connector. Their designators could then automatically be referenced, even for auto-generated ones.

I totally agree. It has been on my mind many times since I wrote my suggestion above, but I haven't given priority to diving into the graphviz package to find how to inject my tweak code for each node. Todays tweak code is very simple using search and replace on the total `.gv` text output from the graphviz package. To do it per node, it will probably require injecting my code earlier in the process, while each node is generated.

> 
> Note that it is possible to define multiple subgraphs with the same name. These will be merged by Graphviz, which allows appending designators to clusters.

That I haven't tried. That sounds like very useful in combination with tweak per node. I have been thinking of adding my own per node specifyer for being included into a common group, but what you describe might be a lower hanging fruit.

_Originally posted by @kvid in https://github.com/wireviz/WireViz/issues/270#issuecomment-2116041655_

kvid:
- Implements feature suggested in #349
- Please suggest a tweak example we could consider adding
- I suggest a squash merge

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -85,6 +85,10 @@ tweak:  # optional tweaking of .gv output
   # loops
   loops: <List>  # every list item is itself a list of exactly two pins
                  # on the connector that are to be shorted
+
+  # optional tweaking of .gv output executed for each instance of this connector
+  tweak:  # see below
+
 ```
 
 ## Cable attributes
@@ -147,6 +151,9 @@ tweak:  # optional tweaking of .gv output
   show_name: <bool>         # defaults to true
   show_wirecount: <bool>    # defaults to true
   show_wirenumbers: <bool>  # defaults to true for cables; false for bundles
+
+  # optional tweaking of .gv output executed for each instance of this cable
+  tweak:  # see below
 
 ```
 
@@ -450,6 +457,13 @@ Alternatively items can be added to just the BOM by putting them in the section 
   # Optional tweaking of the .gv output.
   # This feature is experimental and might change
   # or be removed in future versions.
+
+  placeholder: <str>  # Substring to be replaced with node name
+    # An empty string as placeholder disable replacements.
+    # When placeholder is absent, the global placeholder is used.
+    # For tweak sections in connectors and cables, all substrings
+    # matching the placeholder text will be replaced with the name
+    # of connector/cable in all override and append entries.
 
   override:  # dict of .gv entries to override
     # Each entry is identified by its leading string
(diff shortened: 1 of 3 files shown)


kvid:
See my updated https://github.com/wireviz/WireViz/issues/270#issuecomment-1029535877 for a use case with example YAML input.

martinrieder:
Great to see this implemented! Now I understand how you intended to do that search & replace, though I still do not understand what the placeholder would be needed for. I think that it might make sense to have a prefix to an auto-generated designator though... What is the advantage of having a user-defined ID here?

I also realized that the templates and examples do not contain any tweaks. The automated checks would only confirm that this PR does not break any of them.

martinrieder:
I wrote:
> I also realized that the templates and examples do not contain any tweaks. The automated checks would only confirm that this PR does not break any of them. 

Would it make sense to include some additional examples for these tweaks with the PR? 
In case that you want to leave them undocumented (refer to Graphviz docs), should there be an additional testing folder for the GH actions to capture this?

kvid:
@martinrieder wrote:
> [...]
> > I also realized that the templates and examples do not contain any tweaks. The automated checks would only confirm that this PR does not break any of them.

True

> 
> @kvid : Would it make sense to include some additional examples for these tweaks with the PR, i.e. a simplified version of #325?

It would perhaps be nice with an example, but bear in mind that tweaks are mainly needed for workarounds when the ordinary functionality is not sufficient or contains errors. Hopefully, such needs will be resolved by extending the ordinary functionality in the future.

> In case that you want to leave them undocumented (refer to Graphviz docs),

Creating tweaks does require knowledge of the Graphviz syntax, and that is a too large subject to cover in the WireViz documentation, so referring to Graphviz docs is needed.

> should there be an additional testing folder for the GH actions to capture any issues?

#251 will add test folders that might be what you seek.

> 
> Taking the latter thought a bit further, I come to think of it as regression testing for solved issues... Quite a bit of work to implement though.

See #63

martinrieder:
> > In case that you want to leave them undocumented (refer to Graphviz docs),
> 
> Creating tweaks does require knowledge of the Graphviz syntax, and that is a too large subject to cover in the WireViz documentation, so referring to Graphviz docs is needed.
> 

I would put some notes into syntax.md and amend those to my PR #386 then. We have seen a few requests about clustering. That is something that I would still like to add, besides the hint about the image attributes.

Crestless8:
Being able to do an invisible connector and using autogenerate would be greatly appreciated. The only way it currently works is to enumerate each virtual connector explicitly which is clunky and adds a lot of code.

Issue #355, opened 2024-05-21
[bug] ValueError: X is not in list


the-hess:
First of all thanks everyone who's worked on this, it's a great tool.

I'm having an issue when using arrows and simple connectors.

At the moment I'm trying to do a large wiring diagram for a full system and I'm getting an error when pointing a simple connector to a normal connector with an arrow. It works when I point the connector to pin 1, but not pin 2 or higher.

In my yaml file, the offending entry is on line 149 - if you change the [2] to a [1] the diagram renders as below, but I want it to point to contact 2 of connector 07-0613.

I'm currently using version the 0.4 and have (manually) implemented the basic fixes in #346 and #347.

Graphviz version 9.0.0 (20230911.1827)
Python 3.12.2

![wiringDiag](https://github.com/wireviz/WireViz/assets/49611208/e8ab3b0b-4278-410d-a485-ec7cb88e3d93)

```
options:
  mini_bom_mode: true
  
connectors:
#Items
 07-0601:
    pinlabels: [J14,J15,J28,J4,J10,J12,J5,J7,J1,J6,J8,J2,J13,J16,J11,J9]
    show_pincount: false
    #bgcolor: OL
    
 07-0620:
    pinlabels: [9 Way, 25 Way, 37 Way, 4 Way XLR]
    show_pincount: false

 07-0518:
    pins: [L,N,E,+V,-V]
    show_pincount: false
 
 07-0624: 
    pinlabels: [1]
    show_pincount: false
    
 07-0613:
    pins: [1, 2, 3]
    pinlabels: [J1, J2, J3]
    show_pincount: false
#End of wire points 
 F:
    style: simple
    type: bootlace ferrule
    
 P:
    style: simple
    type: Molex
    
 E:
    style: simple
    type: E-Stop
    
 X:
    style: simple
    type: XLR
    
 D:
    style: simple
    type: D-SUB
    
cables:
 07-0608:
    show_wirecount: false
    colors: [BU,BUWH]
 
 07-0607:
    show_wirecount: false
    colors: [RD,WH,RDWH]
    
 07-0603:
    colors: [RD,BK]
    show_wirecount: false
    
 07-0606:
    colors: [VT,VTWH,VTWH,GN,GNWH,GNWH,OG,OGWH,OGWH]
    show_wirecount: false

 07-0604:
    colors: [GY]
    show_wirecount: false
    
 07-0609:
    colors: [GY]
    show_wirecount: false
    
 07-0610:
    colors: [GY]
[...]

PR #358, opened 2024-05-22
Fix bugs in mate processing

kvid:
The code processing mates used a mix of repeated connector look-ups and local connector variables, and one variable was used before it was assigned the correct value.

This error was detected when investigating the #355 bug report.

- [x] The local connector variables are now both assigned initially when processing each mate, and used instead of repeated connector look-ups.
- [x] The same code section also had an unfortunate side-effect - observed in https://github.com/wireviz/WireViz/issues/325#issuecomment-2116395221
that caused some generated dot output to be unavailable for tweaking. A fix for that is now pushed as a separate commit into this PR.

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -595,6 +595,7 @@ class Harness:
                 typecheck("tweak.append", self.tweak.append, str)
                 dot.body.append(self.tweak.append)
 
+        # TODO: All code below until "return dot" must be moved above all tweak processing!
         for mate in self.mates:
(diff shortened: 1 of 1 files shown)


amotl:
Reviewing this patch may need your attention, @formatc1702. Do you have a few spare cycles?

kvid:
@formatc1702 - Please review, but not merge yet this PR.

formatc1702:
Set to draft to prevent premature merge as requested :)

formatc1702 (approved):
Good fix, thanks!

Issue #227, opened 2021-03-23
[feature] Include BOM table in graphical output


formatc1702:
GraphViz allows HTML tables in a `graph`'s `label`:

```
graph {
	graph [bgcolor=white fontname=arial label=<<table border="0" cellspacing="0" cellpadding="3" cellborder="1"><tr><td>A</td><td>B</td></tr><tr><td>C</td><td>D</td></tr></table>>]
}

// etc.
```

<img width="920" alt="image" src="https://user-images.githubusercontent.com/2447251/112132865-5d9b1a80-8bcb-11eb-8ea8-2c694de75314.png">

This could be used to optionally include the complete BOM table inside the graphical output (PNG/SVG) if desired.

PR #367, opened 2024-05-29
Added final stuff for PDF output

tobiasfalk:
Everythnig was alredy prepairt and I just had to make some minor changes. I do not know why this was prepaired but not finished.


martinrieder (approved):
As long as you do not need drawing frames and BOM, this is fine.

tobiasfalk:
@martinrieder yes and no, Yes it does not add frames and so. And No since I originally just wanted a PDF output and was not interested in the frame, since PDF is just a robuster vector graphics format and is less dependent on the interpreter.

martinrieder:
Maybe we need to make some destinguishment through another CLI argument that would request BOM output, templates and metadata to be included, which is currently only supported for HTML.

kvid:
@martinrieder wrote:
> Maybe we need to make some destinguishment through another CLI argument that would request BOM output, templates and metadata to be included, which is currently only supported for HTML.
> 
> EDIT: Alternatively, the list of output formats for the `-f` CLI switch might be extended to contain PDF either with or without BOM.

If you want to include a BOM in the diagram (as in #227) and optionally also a frame, then such extra elements will be added to all graphical output types, and controlling this by different output type specifiers makes no sense to me. A common specifyer for what to include in the diagram is then needed, and the HTML output should be adjusted accordingly to avoid doubled BOM and frame.

PR #379, opened 2024-06-08
Added option to set the dpi of the output

tobiasfalk:
This adds the option 'output_dpi' for the user. This is used to set the 'dpi' of the output, the default is 96 which is also the default GraphViz default. https://graphviz.org/docs/attrs/dpi/

This comes from my annoyance that the text is kind of pixelated with 96dpi and having the option for higher res pictures is nice


tobiasfalk:
@kvid, since someone asked about it, what about merging this?

kvid:
> @kvid, since someone asked about it, what about merging this?

Personally, I would consider a solution that also allows overriding some other global parameters, e.g. something like this:
```yaml
options:
  graph:
    dpi: 600
    ranksep: 4
    ... etc.
```
However, I've not tested this, so please argue against me and explain why my idea should not be implemented.

PR #380, opened 2024-06-09
Rename placehoder "sheetsize_default" to <!-- %template_sheetsize% -->

### LINKED ISSUE #377: [feature] Unfortunate HTML output template placeholder
HTML output templates with placeholders for generated text is a new feature introduced in v0.4. All placeholders are named `<!-- %something% -->`, except one placeholder called `"sheetsize_default"` only.
1. Why is it called `sheetsize_default` and not `sheetsize` or `template_sheetsize`? The replacement text is fetched from `metadata.template.sheetsize`.
2. Why is it not called `<!-- %template_sheetsize% -->` to be consistent with the other placeholders?
3. I guess the quote characters was needed to include in the search term to avoid false hits, and therefore re-inserted around the fetched value to keep the correct HTML syntax. This seems like a hack for this placeholder only, and would not be needed if using my suggestion in point 2 above. Then `<!-- %template_sheetsize% -->` could be placed within the quote characters in the template, but the quote characters need not be replaced, and special treatment of this placeholder is no longer needed. The code would be cleaner, and a documentation description would be easier to understand because all placeholders are treated equally.

I suggest replacing https://github.com/wireviz/WireViz/blob/v0.4/src/wireviz/wv_html.py#L104-L106 with a simple entry
```py
        "<!-- %template_sheetsize% -->": metadata.get("template", {}).get("sheetsize", ""),
```
inserted together with the others at https://github.com/wireviz/WireViz/blob/v0.4/src/wireviz/wv_html.py#L88
and then update the placeholder to `<!-- %template_sheetsize% -->` in `templates/din-6771.html`

kvid:
@formatc1702 must must check if this suggested change (placeholder name consistency) is OK, or if it might have some bad side-effects. If there is a good reason to keep this one placeholder very different from the others, then that reason should be explained in a code comment.

A bonus advantage by using the name `<!-- %template_sheetsize% -->`, is that it will be easy to later generalize by letting any string or numeric `metadata.{item}.{key}` entry replace the corresponding `<!-- %{item}_{key}% -->` placeholder.

~TODO: If both PR #371 and this PR is accepted, then the `templates/README.md` file must be updated.~ (it is now updated)

Fixes #377

### DIFF
--- a/src/wireviz/templates/README.md
+++ b/src/wireviz/templates/README.md
@@ -43,6 +43,7 @@ Note that there must be one single space between `--` and `%` at both ends.
 | `<!-- %{item}% -->`           | String or numeric value of `metadata.{item}` |
 | `<!-- %{item}_{i}% -->`       | Category number `{i}` within dict value of `metadata.{item}` |
 | `<!-- %{item}_{i}_{key}% -->` | Value of `metadata.{item}.{category}.{key}` |
+| `<!-- %template_sheetsize% -->` | Value of `metadata.template.sheetsize` |
 
 Note that `{item}`, `{category}` and `{key}` in the description above can be
 any valid YAML key, and `{i}` is an integer representing the 1-based index of
--- a/src/wireviz/templates/din-6771.html
+++ b/src/wireviz/templates/din-6771.html
@@ -179,7 +179,7 @@
 </head>
 <body>
   <div id="page">
-    <div id="frame" class="sheetsize_default">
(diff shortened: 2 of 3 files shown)


formatc1702:
Please consider my [comment on the original issue](https://github.com/wireviz/WireViz/issues/377#issuecomment-2161493538), but generally I am all in favor of making placeholder syntax consistent and therefore, of amending the Readme (please do) and merging this PR.

formatc1702 (approved):

Issue #342, opened 2024-05-08
Windows build failing


VIPQualityPost:
Hi,

I installed on macos, it works fine. I tried to install on my work windows machine today, and am unable to generate even demo examples from repo: 
```
metadata = Metadata(**yaml_data.get('metadata', {})),
AttributeError: 'NoneType' object has no attribute 'get'
```

VIPQualityPost:
This was with demo02.yml from the repo.

kvid:
@VIPQualityPost - thank's for reporting this error, but please provide information about your set of versions as described in [CONTRIBUTING.md](https://github.com/wireviz/WireViz/blob/master/docs/CONTRIBUTING.md) and how you installed WireViz and dependencies. Otherwise, it might be hard to help you when nobody can recreate the error you report.

kvid:
@VIPQualityPost wrote:
> [...]
> ```
> metadata = Metadata(**yaml_data.get('metadata', {})),
> AttributeError: 'NoneType' object has no attribute 'get'
> ```

The error message you report tells me that you have no `yaml_data`, which is highly unusual. Either your YAML input is very weird (which isn't the case if you use the original "demo02.yml from the repo"), or maybe something's wrong with your `PyYAML` library installation, but I need more information from your installation setup to know.

Please try my suggestion in https://github.com/wireviz/WireViz/issues/366#issuecomment-2135646682 and report here the full output from all these commands.

VIPQualityPost:
Sorry about disappearing off the issue- have been moving and not had access to my windows machine in the meantime. I will grab that info and get you more detail in a week or two!

miroc99:
I also have the same problem, install wireviz 0.4 from pip and use demo01.yml to test but fail
os: Windows10 22H2
python: 3.12.2(venv)
wireviz: 0.4

`WireViz 0.4
Input file:   demo01.yml
Output file:  demo01.[html|png|svg|tsv]
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "d:\Dev\wireviz\.venv\Scripts\wireviz.exe\__main__.py", line 7, in <module>
  File "d:\Dev\wireviz\.venv\Lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #383, opened 2024-06-11
Explain unexpeced top-level type

kvid:
Might help in use cases like #342


src/wireviz/wireviz.py
```diff
@@ -86,6 +86,9 @@ def parse(
         raise Exception("No output formats or return types specified")
 
     yaml_data, yaml_file = _get_yaml_data_and_path(inp)
+    assert isinstance(
+        yaml_data, dict
+    ), f"Expected a dict/map as top-level YAML input, but got: {type(yaml_data)}"
```
amotl, line 91: In runtime code, it is better to handle type checks this way. `assert` statements are still applicable to have in software tests.

```suggestion
    if not isinstance(yaml_data, dict):
        raise TypeError(f"Expected a dict/map as top-level YAML input, but got: {type(yaml_data)}"
```

amotl (approved):

src/wireviz/wireviz.py
amotl: Thanks!

src/wireviz/wireviz.py
```diff
@@ -86,6 +86,8 @@ def parse(
         raise Exception("No output formats or return types specified")
 
     yaml_data, yaml_file = _get_yaml_data_and_path(inp)
+    if not isinstance(yaml_data, dict):
+        raise TypeError(f"Expected a dict/map as top-level YAML input, but got: {type(yaml_data)}")
```
formatc1702, line 90: What does map mean in this context?

kvid: Some YAML descriptions use "map", others call it "dictionary", and "hash" is also used. In the Python world, I guess "dict" is more common, so I have now changed it to a simple "dict". See e.g. https://www.yaml.info/learn/index.html

PR #386, opened 2024-06-14
Image syntax description missing `scale` and `fixedsize`.

martinrieder:


docs/syntax.md
```diff
@@ -530,6 +530,8 @@ image:
   height: <int>      # range: 1~65535; unit: points
   # if only one dimension (width/height) is specified, the image is scaled proportionally.
   # if both width and height are specified, the image is stretched to fit.
+  scale: <str>     # Which dimension to scale ['false', 'true', 'width', 'height', 'both']
+  fixedsize: <bool>  # If only one dimension is specified, compute the other
 ```
 
 For more fine grained control over the image parameters, please see [`advanced_image_usage.md`](advanced_image_usage.md).
```
kvid, line 537: Please read the whole Images section, and you will understand.

kvid (changes requested):
I'm sorry, I cannot approve this PR.

docs/syntax.md
martinrieder: Understood, though I think having a hint on these additional attributes might be helpful. The `image` description should either be complete or entirely moved into the other file.

kvid: As far as I remember from almost 4 years ago when the image feature was added, @formatc1702 was very concerned about possibly confusing many users with these two attributes. We therefore put a great effort into computing default values for them such that more than 99 % of the use cases would not need to know about them, and added a separate advanced image usage document for the remaining less than 1 %.

However, even if @formatc1702 might have changed his mind since then, and could be convinced to include them here, I advice against any explanation of them other than referring to the advanced image usage document.

martinrieder: Thanks for the detailed explanation. We should put a note into the code instead, because of this:
https://github.com/wireviz/WireViz/issues/348#issuecomment-2167995260
 > Ideally, there should be some functionally in WireViz to generate matching files based on internal data structures:
> 
> - JSON Schema definitions
> - YAML example/templates

formatc1702: > We therefore put a great effort into computing default values for them such that more than 99 % of the use cases would not need to know about them, and added a separate advanced image usage document for the remaining less than 1 %.

correct :) we want to keep it simple.

PR #309, opened 2023-03-08
Python version update

MNiedzielski:
Enable tests for current Python versions.

PR #388, opened 2024-06-15
Update CI Python versions

ferdnyc:
Python 3.7 is past end-of-life, and even 3.8 is pretty outdated. Switch to testing on 3.8 (oldest supported) and 3.12 (current release).


kvid:
Thank you for this contribution, but it seems to be a duplicate of PR #309. Please add your contributions there if you agree.

ferdnyc:
@kvid So it does, my apologies. It's so far down the list I didn't spot it there!

Issue #391, opened 2024-06-19
Help with OS Error


ElPi23:
Hello everyone,
I have received the following error and can't figure out what the problem is.

Error:
`ERROR: Traceback (most recent call last):
File "", line 198, in _run_module_as_main
File "", line 88, in _run_code
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Scripts\wireviz.exe\__main__.py", line 7, in
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 1157, in __call__
return self.main(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 1078, in main
rv = self.invoke(ctx)
^^^^^^^^^^^^^^^^
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 1434, in invoke
return ctx.invoke(self.callback, **ctx.params)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 783, in invoke
return __callback(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\wireviz\wv_cli.py", line 141, in wireviz
wv.parse(
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\wireviz\wireviz.py", line 88, in parse
yaml_data, yaml_file = _get_yaml_data_and_path(inp)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\wireviz\wireviz.py", line 416, in _get_yaml_data_and_path
raise e
File "C:\Users\elias.pircher\AppData\Local\Programs\Python\Python312\Lib\site-packages\wireviz\wireviz.py", line 406, in _get_yaml_data_and_path
yaml_path = Path(inp).expanduser().resolve(strict=True)
[...]

PR #365, opened 2024-05-25
Release v0.4.1

kvid:
- [x] #346 
- [x] #347 
- [x] #358 
- [x] #361 
- [x] #363 
- [x] #364 
- [x] #371 
- [x] #374 
- [x] #375 
- [x] Cherry-pick doc addition https://github.com/wireviz/WireViz/commit/590504130c5e878c51df571974ee31e6f01b3a53 from `dev`
- [x] #380 
- [x] #381 
- [x] #383 
- [x] #392 
- [x] Print Python version & OS information when exception is raised in PR above
- [x] #395 
- [x] ~TODO: other fixes?~ Issue freeze
- [x] Interactive rebase to update PR refs
- [x] Update any missing `CHANGELOG` entries (remember to use new URL)
- [x] Apply `isort` & `black` on all `.py` files
-----
Should we avoid pushing these last steps to this PR branch? Or should they be pushed together just before PR merge into `master` and creating the release? Anything missing?
- [x] Bump version to v0.4.1
- [x] Execute `build_examples.py` and commit generated files
- [x] Update release date in CHANGELOG 
- [ ] Create release and publish it
-----
After the release:
- Apply all release changes to `dev` after or together with #251 merge-in to avoid regressions (manage in separate new PR!)

To try these fixes before release, this is one method:
```sh
pip uninstall wireviz
pip install git+https://github.com/wireviz/WireViz@release/v0.4.1-rc
```

PR #346, opened 2024-05-15
Handle OSError [errno.EINVAL] that might be raised in Windows

kvid:
Suggested fix posted in
https://github.com/wireviz/WireViz/issues/344#issuecomment-2113476151

Please verify that this fixes #344 and report back.

~~The base branch of this PR is `master` for the time beeing, but it should be changed to a new release/v0.4.1-rc before any merge-in. #345 should also be fixed and updating the changelog before releasing the hotfix.~~

Issue #344, opened 2024-05-13
[bug] OSError: [WinError 123] with version 0.4


KlopDesign:
Thanks for the great tool! However, with version WireViz 0.4 I cannot get a good output. 

With `wireviz demo01.yml` and `wireviz tutorial01.yml` I get: 
`OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: ...`
Python version 3.7.0 using [pyenv](https://github.com/pyenv-win/pyenv-win). 

I installed WireViz 0.3.2 and that one works just fine. However, demo02.yml does not work as in #304: Exception: First item not found anywhere.

Edit 14-05-2024: hereby my cmd output. Installed Python via pyenv: `pyenv install 3.7.0`
```
Microsoft Windows [Version 10.0.19045.4291]
(c) Microsoft Corporation. Alle rechten voorbehouden.

E:\####\Downloads>pyenv global 3.7.0

E:\####\Downloads>pip3 install wireviz
Collecting wireviz
  Using cached wireviz-0.4-py3-none-any.whl.metadata (6.1 kB)
Requirement already satisfied: click in c:\users\####\.pyenv\pyenv-win\versions\3.7.0\lib\site-packages (from wireviz) (8.1.7)
Requirement already satisfied: pyyaml in c:\users\####\.pyenv\pyenv-win\versions\3.7.0\lib\site-packages (from wireviz) (6.0.1)
Requirement already satisfied: pillow in c:\users\####\.pyenv\pyenv-win\versions\3.7.0\lib\site-packages (from wireviz) (9.5.0)
Requirement already satisfied: graphviz in c:\users\####\.pyenv\pyenv-win\versions\3.7.0\lib\site-packages (from wireviz) (0.20.1)
Requirement already satisfied: colorama in c:\users\####\.pyenv\pyenv-win\versions\3.7.0\lib\site-packages (from click->wireviz) (0.4.6)
Requirement already satisfied: importlib-metadata in c:\users\####\.pyenv\pyenv-win\versions\3.7.0\lib\site-packages (from click->wireviz) (6.7.0)
Requirement already satisfied: zipp>=0.5 in c:\users\####\.pyenv\pyenv-win\versions\3.7.0\lib\site-packages (from importlib-metadata->click->wireviz) (3.15.0)
[...]

PR #392, opened 2024-06-20
Fix Exception Handling in _get_yaml_data_and_path(...)

martinrieder:
Add ValueError and OSError where `e.errno` is `None` to Exception Handling of `_get_yaml_data_and_path()`

**This PR is amending fixes for the following issues based on #250 and #318.**
- https://github.com/wireviz/WireViz/pull/318#pullrequestreview-1457016602
  describing 80000 character YAML input raising `ValueError`
- https://github.com/wireviz/WireViz/issues/391#issuecomment-2180419249 
  describing `OSError: _getfinalpathname: path too long for Windows` 

```
# if inp is a long YAML string, Pathlib will raise OSError [errno.ENAMETOOLONG]
# in Windows, ValueError or OSError [errno.EINVAL or None] also might be raised
# when trying to expand and resolve it as a path (depending on the Python version)
# Catch these specific errors, but raise any others
```
 
`master` contains #250 from d7d7854bce39881a970b54e3522d9293441c5ce1: Consolidate wireviz.parse() to handle Path, str and Dict as input

https://github.com/wireviz/WireViz/blob/954c4f5f92b1d0531f6af0c318c60786b7906d42/src/wireviz/wireviz.py#L415

`release/v0.4.1-rc` contains #318 at 6f9007f45d48cc65f3469b362976bb53ac1cd7e0: Use portable OS error codes so program doesn't crash

https://github.com/wireviz/WireViz/blob/ba84c54382b540dfa50dcd5e3592917abdbfb1f8/src/wireviz/wireviz.py#L419

Due to the following https://github.com/wireviz/WireViz/pull/246#issuecomment-1206635845, this is a bugfix only, waiting to be resolved by #251. 
> All comments have been addressed inside #251 or are no longer relevant after 251 is applied

_References_
- #365
  - #391
    #392
  - #344
    #346
  - #318
- #190
  #246
- #250
  #251

### DIFF
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -410,10 +410,11 @@ def _get_yaml_data_and_path(inp: Union[str, Path, Dict]) -> (Dict, Path):
             # if no FileNotFoundError exception happens, get file contents
             yaml_str = open_file_read(yaml_path).read()
         except (FileNotFoundError, OSError, ValueError) as e:
-            # if inp is a long YAML string, Pathlib will raise OSError: [errno.ENAMETOOLONG]
-            # (in Windows, it seems OSError [errno.EINVAL] might be raised in some cases)
-            # when trying to expand and resolve it as a path.
-            # Catch this error, but raise any others
+            # if inp is a long YAML string, Pathlib will raise OSError [errno.ENAMETOOLONG]
+            # in Windows, ValueError or OSError [errno.EINVAL or None] also might be raised
+            # when trying to expand and resolve it as a path (depending on the Python version)
+            # Catch these specific errors, but raise any others
+
             from errno import EINVAL, ENAMETOOLONG
 
             if type(e) is OSError and e.errno not in (EINVAL, ENAMETOOLONG, None):
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -416,7 +416,7 @@ def _get_yaml_data_and_path(inp: Union[str, Path, Dict]) -> (Dict, Path):
             # Catch this error, but raise any others
             from errno import EINVAL, ENAMETOOLONG
 
-            if type(e) is OSError and e.errno not in (EINVAL, ENAMETOOLONG):
+            if type(e) is OSError and e.errno not in (EINVAL, ENAMETOOLONG, None):
                 raise e
             # file does not exist; assume inp is a YAML string
             yaml_str = inp
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -410,10 +410,12 @@ def _get_yaml_data_and_path(inp: Union[str, Path, Dict]) -> (Dict, Path):
             # if no FileNotFoundError exception happens, get file contents
             yaml_str = open_file_read(yaml_path).read()
         except (FileNotFoundError, OSError, ValueError) as e:
-            # if inp is a long YAML string, Pathlib will raise OSError [errno.ENAMETOOLONG]
-            # in Windows, ValueError or OSError [errno.EINVAL or None] also might be raised
-            # when trying to expand and resolve it as a path (depending on the Python version)
-            # Catch these specific errors, but raise any others
+            # if inp is a long YAML string, Pathlib will normally raise
+            # FileNotFoundError or OSError(errno = ENAMETOOLONG) when
+            # trying to expand and resolve it as a path, but in Windows
+            # might ValueError or OSError(errno = EINVAL or None) be raised
+            # instead in some cases (depending on the Python version).
+            # Catch these specific errors, but raise any others.
 
             from errno import EINVAL, ENAMETOOLONG
 
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -409,7 +409,7 @@ def _get_yaml_data_and_path(inp: Union[str, Path, Dict]) -> (Dict, Path):
             yaml_path = Path(inp).expanduser().resolve(strict=True)
             # if no FileNotFoundError exception happens, get file contents
             yaml_str = open_file_read(yaml_path).read()
-        except (FileNotFoundError, OSError) as e:
+        except (FileNotFoundError, OSError, ValueError) as e:
             # if inp is a long YAML string, Pathlib will raise OSError: [errno.ENAMETOOLONG]
             # (in Windows, it seems OSError [errno.EINVAL] might be raised in some cases)
             # when trying to expand and resolve it as a path.


kvid:
@martinrieder wrote:
> I noticed that a similar issue has been found by @kvid in [#318 (review)](https://github.com/wireviz/WireViz/pull/318#pullrequestreview-1457016602) a year ago. Could you please check this again?
> 
> It seems that behavior changes with different Python versions. The referred one uses `ValueError` in 3.9, while it is now reported as `OSError` in 3.12.

I can confirm that I still get `ValueError: _getfinalpathname: path too long for Windows` when trying WireViz on input YAML with 80000 character dummy string in Python 3.10.4 at Windows 10. It's not a very likely use case, but demonstrates a possible exception when trying to use random strings as a path. This edge case is resolved by including `ValueError` in the list of exceptions to catch.

kvid:
> what does errno evaluate to in that case?

An `OSError` exception has `errno`, but `ValueError` has not.

> Which version of Pathlib are you running?

`pathlib` is a system library included in (and identified by) the Python 3.10.4 installation in my case. System libraries doesn't have separate version numbers AFAIK.

> 
> BTW: I do not consider running a huge YAML file to be that unusual. If a library of connectors is contained in the templates section, file size can increase quickly.

I agree to that, but my 80000 character test input was only a single long string, and that is a bit unusual. However, I just now added my 80000 character as one of the attributes of a connector in a valid WireViz YAML file, and the same ValueError was raised, so you are probably right it also can happen with a huge template section.

kvid:
> > I suggest you include `ValueError` like this:
> > ```python
> >         except (FileNotFoundError, OSError, ValueError) as e:
> > ```
> 
> Additionally, the if clause needs to be adapted, so that the exception is also raised.

No, the `if` only needs to raise other `OSError` exceptions than the identified `errno` values.

> 
> My proposal to add some "band-aid" the fix:
> 
> * **If `inp` contains newline characters, treat it as YAML**

As I wrote a year ago in point 4 of https://github.com/wireviz/WireViz/pull/318#pullrequestreview-1457016602 that you referred to above, this is exactly what is already implemented in #251.

Catching more exceptions here is just a quick band-aid to avoid most errors until the more thorough fix in #251.

> * Else pass it to Pathlib and catch its exceptions
> * Raise any unexpected exceptions

No expected exceptions to catch after the #251 changes, so they are all raised normally.

> * Fall back to YAML parsing on FileNotFoundError etc.

No special catching of these exceptions are needed.

> 
> What do you think about treating all multi-line strings as YAML?

I agree, and #251 is the first to be merged after releasing the collections of quick-fixes.

> 
> Besides bugfixing, would printing an error message if it is neither an existing file nor valid YAML provide a more user-friendly behavior instead of simply raising an exception?

See #207

kvid:
@formatc1702 seems to prefer not referring to issues or PRs in the source code - see https://github.com/wireviz/WireViz/pull/264#discussion_r1319052618

Referring to them in the body of commit messages is OK and often useful, though.

I therefore suggest removing the two trailing comments including issue references. Instead I suggest expanding the comment block to explain the latest additions like this:
```py
    # if inp is a long YAML string, Pathlib will raise OSError: [errno.ENAMETOOLONG]
    # (in Windows, it seems ValueError or OSError [errno.EINVAL or None]
    # also might be raised in some cases - depending on the Python version)
    # when trying to expand and resolve it as a path.
    # Catch these errors, but raise any others
```

kvid:
@martinrieder wrote:
> Thanks. I will update the PR accordingly.
> 
> I discovered an additional OS-specific bug with `expanduser()` that arises when the file name contains a tilde `~` character. Should I include a fix for this as well, even though it is a separate issue?

If the fix is similar to what is already discussed (catching an extra `Exception` class or `OSError.errno` value), then I see no harm in including also that fix in this PR - just update the PR title and description to cover all sub-issues.

Please [use the imperative mood in your commit subject lines](https://cbea.ms/git-commit/#imperative) as recommended in the [contribution guidelines](https://github.com/wireviz/WireViz/blob/v0.4/docs/CONTRIBUTING.md#hints) and I think the PR title should follow the same recommendation as for commit subject lines, because a squash commit will normally use the PR title as subject line.

kvid:
I'm sorry for being unclear. When I wrote "_update the PR title and description to cover all sub-issues_", I didn't mean all details should be in the title. Please read all the commit recommendations at the page I linked to above. The subject line should be a short summary, preferably limited to 50 characters and never more than 72. Use the body for details.

formatc1702:
Thanks for digging into these more obscure issues despite my limited availability. It seems like a very minor code change with multiple positive effects, good job :) I assume it's ready for merge but have requested review just to be sure.

kvid:
@martinrieder wrote:
> [...]
> I discovered an additional OS-specific bug with `expanduser()` that arises when the file name contains a tilde `~` character. Should I include a fix for this as well, even though it is a separate issue?

Didn't you include a fix for this? I can't see any more code additions than what we discussed before you mentioned this last issue.

Or does the fix involve huge code changes that is clearly out of scope for this PR?

kvid:
At Linux it's normal to evaluate a path with a leading `~Martin` to the home folder of a user named `Martin`, and it seems Python tries to make this work in a similar fashion at Windows. Se also https://docs.python.org/3/library/os.path.html#os.path.expanduser
and https://docs.python.org/3/library/pathlib.html#pathlib.Path.expanduser

If I understand your description correctly, it seems to work as designed. You get an exception because there is no user named `demo01.yml` at your PC, and hence no home folder named `C:\\Users\\demo01.yml`.

src/wireviz/wireviz.py
```diff
@@ -409,14 +409,15 @@ def _get_yaml_data_and_path(inp: Union[str, Path, Dict]) -> (Dict, Path):
...
-        except (FileNotFoundError, OSError) as e:
-            # if inp is a long YAML string, Pathlib will raise OSError: [errno.ENAMETOOLONG]
-            # (in Windows, it seems OSError [errno.EINVAL] might be raised in some cases)
-            # when trying to expand and resolve it as a path.
-            # Catch this error, but raise any others
+        except (FileNotFoundError, OSError, ValueError) as e:
+            # if inp is a long YAML string, Pathlib will raise OSError [errno.ENAMETOOLONG]
+            # in Windows, ValueError or OSError [errno.EINVAL or None] also might be raised
+            # when trying to expand and resolve it as a path (depending on the Python version)
+            # Catch these specific errors, but raise any others
```
kvid, line 416: I feel your latest rewriting of these comments might have unintentionally changed the meaning a bit, so I suggest a new rewriting that hopefully is easier to understand:
```suggestion
            # if inp is a long YAML string, Pathlib will normally raise
            # FileNotFoundError or OSError(errno = ENAMETOOLONG) when
            # trying to expand and resolve it as a path, but in Windows
            # might ValueError or OSError(errno = EINVAL or None) be raised
            # instead in some cases (depending on the Python version).
            # Catch these specific errors, but raise any others.
```

kvid (changes requested):
- `build_examples.py` runs without errors, and produces the expected results (diffs on version numbers and image absoulute paths are normal).
- My `a-80k.yml` from https://github.com/wireviz/WireViz/pull/392#issuecomment-2192442165 is processed without errors, and produces the expected result.
- I'm not able to verify the fix for #391 because I'm not able to reproduce the reported error with `YMLFILE.txt` from #391 (I've tried with Python 3.9.13 and 3.12.4 because 3.12.3 is no longer available from Microsoft Store). I get no errors both before and after changes in this PR.
- I suggest a simple squash and merge with a summary of the two resolved issues in the resulting commit body.

kvid (approved):

PR #395, opened 2024-06-23
Avoid ResourceWarning: unclosed file

kvid:
A number of such warnings showed up when running (with wireviz v0.4) e.g.
```sh
PYTHONWARNINGS=always python build_examples.py
PYTHONWARNINGS=always wireviz ../../examples/demo0?.yml
```
See https://github.com/wireviz/WireViz/pull/309#issuecomment-2170988381

Fix: All `open()` calls should be in a "`with open() as x`" statement to ensure closing the file when exiting the block in any way. Otherwise, use the new `file_read_text()` or `file_write_text()` thin wrapper functions to read or write the whole utf-8 text file and closing it.

### DIFF
--- a/src/wireviz/wv_html.py
+++ b/src/wireviz/wv_html.py
@@ -35,11 +35,11 @@ def generate_html_output(
         # fall back to built-in simple template if no template was provided
         templatefile = Path(__file__).parent / "templates/simple.html"
 
-    html = file_read_text(templatefile)
+    html = file_read_text(templatefile)  # TODO?: Warn if unexpected meta charset?
 
     # embed SVG diagram (only if used)
     def svgdata() -> str:
-        return re.sub(
+        return re.sub(  # TODO?: Verify xml encoding="utf-8" in SVG?
             "^<[?]xml [^?>]*[?]>[^<]*<!DOCTYPE [^>]*>",
             "<!-- XML and DOCTYPE declarations from SVG file removed -->",
             file_read_text(f"{filename}.tmp.svg"),
--- a/src/wireviz/wv_html.py
+++ b/src/wireviz/wv_html.py
@@ -9,9 +9,9 @@ from wireviz.DataClasses import Metadata, Options
 from wireviz.svgembed import data_URI_base64
 from wireviz.wv_gv_html import html_line_breaks
 from wireviz.wv_helper import (
+    file_read_text,
+    file_write_text,
     flatten2d,
-    open_file_read,
-    open_file_write,
     smart_file_resolve,
 )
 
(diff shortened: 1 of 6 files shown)


src/wireviz/wv_html.py
```diff
@@ -35,14 +35,14 @@ def generate_html_output(
...
-    html = open_file_read(templatefile).read()
+    html = file_read_text(templatefile)
 
     # embed SVG diagram (only if used)
     def svgdata() -> str:
         return re.sub(
             "^<[?]xml [^?>]*[?]>[^<]*<!DOCTYPE [^>]*>",
             "<!-- XML and DOCTYPE declarations from SVG file removed -->",
-            open_file_read(f"{filename}.tmp.svg").read(),
+            file_read_text(f"{filename}.tmp.svg"),
```
kvid, line 45: This question might be slightly out of scope for this PR, but needs to be raised: Here, the SVG file is assumed to be utf-8 encoded, but it's not assumed at:
https://github.com/wireviz/WireViz/blob/close_files/src/wireviz/svgembed.py#L62-L64

What is correct? Should we assume utf-8 at both places, or does it really depend on what encoding is specified in the leading part of the SVG file itself?

formatc1702: Good question. Since the HTML template contains `<meta charset="UTF-8">`, it would be best if the embedded SVG also was UTF-8. Not sure if we know or have control over what encoding Graphviz chooses to output.

kvid: We probably should assume utf-8 and specify that encoding at both places. I also found a third place where that is already done:  https://github.com/wireviz/WireViz/blob/close_files/src/wireviz/Harness.py#L662

Then, we probably also should (as a minimum only where embedding SVG into HTML, or ideally at all places reading SVG) verify that the `encoding` property of the leading `xml` tag is either absent (utf-8 is default, I believe) or equal to any of the legal value variations that specify utf-8? If we detect a discrepancy, should we raise an exception or just print a warning that some characters might be rendered wrongly due to an unexpected `encoding` in the SVG file?

formatc1702: TBH, I would prefer to keep it simple here and work with the assumption, in order to finish the release of v4.1.
Verification can be done as a separate feature/PR.

kvid: I created #400 to follow up this issue. Please add your concerns and opinions there.

formatc1702: OK thanks. I see some recent force-pushes so please re-request review once it's ready :)

formatc1702 (approved):

PR #398, opened 2024-06-30
Add options.terminology to override certain terms

### LINKED ISSUE #331: [feature] Option to change connector terminology "pin", to "way" or customise
Hi, the word "pin" implies a male contact connector. 

"Pin" is used commonly for through-hole PCB parts, as these will always be male gender, but is less relevant to connectors for wiring harnesses which can be either gender, I shall elaborate:

In this example, the connectors all state "4-pin", but the two connectors on the right are female, and actually don't have round pin-shaped contacts, but sliding contacts similar to a 3.5mm headphone jack: https://scottbouch.com/tmp/UHF2.html

A de-gendered approach is to use the word "way" to describe the number of electrical paths through the connector, and then specify contact gender separately, ie; 

6 way male contact, fixed connector = 6 way plug, panel mounted
6 way female contact, free connector = 6 way socket, cable mounted

In these two examples above, the terms "plug" and "socket" define the gender of the contacts within the connector, and the terms "fixed" and "free" define the mounting of the connector bodies.

- The term "Way" is commonly used to describe the number of electrical signals the connector can carry ("pin" implies the gender of a contact).
- The term "Contact" is commonly used to physically describe the presence of metallic parts used, where a gender can be applied.
- The terms "Male" and "Female" are commonly used to determine the gender of the contact, if relevant. "Male" = plug, "Female" = socket, either can be free of fixed.
- Sometimes contacts are replaced by a plastic plug too, this is where not all contacts are needed, but the IP ingress protection needs to be maintained.
- The term "Fixed" is commonly used to mean "panel mounted".
- The term "Free" is commonly used to mean "cable mounted".

Some people will be happy to stay with the term "Pin" as it is used a lot in language, but it would be a real improvement to add an option to change the displayed term to "Way" or a custom defined term perhaps.

Many thanks, Scott.

kvid:
A selection of terms can optionally be replaced by user defined terms, e.g. like this:
```yaml
options:
  terminology:
    pin: way
    wire: core
    shield: screen
```
Resolves #331 basic feature


amotl (approved):

kvid:
@amotl - Thank's for approving, but it still seems to be some unsettled issues at #331, so I marked this PR a draft until we agree on what to do.

Issue #376, opened 2024-06-05
[feature] Single wire cable with `style: simple`


martinrieder:
The following example brings up an interesting use case for cables containing only a single wire. Similar to the way that connectors can be defined with `style: simple`, I suggest to implement cables with this attribute. It would not be displayed as a box and should optionally show the wire name/label. 

_ Credits to @EloiVilalta in https://github.com/wireviz/WireViz/issues/293#issuecomment-2150401138_

```yaml
connectors:
  C1:
    style: simple
    type: Connector
  C2:
    style: simple
    type: Connector      
cables:  
  W1:
    wirecount: 1
    show_name: false
    show_wirecount: false
    colors: [RD]

connections:
  -
    - C1: [1]
    - W1: [1]
    - C2: [1]
``` 

The issue with the current implementation is that it requires a lot of space in the graph due to the HTML table. This is not needed for only displaying a the wire label/name as a text. 

![image](https://github.com/wireviz/WireViz/assets/32335173/74517787-c508-43d7-8bac-a23e602dea0b)

martinrieder:
@kvid I noticed that you discussed some similar idea with @formatc1702 a few years back in #31. Do you have any updates on the topic? 

> Another thing that might help, is to combine all invisible nodes on the same rank inside a sheathing into a common invisible node to avoid that Graphviz swap their order.

Taking the picture by @SnowMB as an example from this:
> You can use `rank=same` to clean up node placement. 
> 
> ![graph(2)](https://user-images.githubusercontent.com/7050959/88187486-e0cc3100-cc36-11ea-8aa7-c475cb708be4.png)
>

kvid:
See also https://github.com/wireviz/WireViz/issues/225#issuecomment-951164193 where @formatc1702 strongly considers consolidating the optional `style` attribute into `category` that already exists for both connectors and cables.

Maybe this issue is an extra argument for such a consolidation?

martinrieder:
If connector `type:` would not already be in use, I would suggest to use this instead of `category:` for both cables and connectors.

kvid:
@martinrieder wrote:
> I like the idea of consolidating. If connector `type:` would not already be in use, I would suggest to use this instead of `category:` for both cables and connectors.

When consolidating, we might end up with allowing more than one value for the same connector/cable, and if so, then `flags` or `class` might be alternative names? I suggest we try collecting different values that might be useful in a consolidated attribute, and see if they are mutually exclusive.

[...]

PR #405, opened 2024-07-13
[WIP] Implement single wire cable

martinrieder:
DRAFT implementation of
- #376

### DIFF
--- a/src/wireviz/DataClasses.py
+++ b/src/wireviz/DataClasses.py
@@ -271,6 +271,7 @@ class Cable:
     show_name: Optional[bool] = None
     show_wirecount: bool = True
     show_wirenumbers: Optional[bool] = None
+    show_colorname: bool = True
     ignore_in_bom: bool = False
     additional_components: List[AdditionalComponent] = field(default_factory=list)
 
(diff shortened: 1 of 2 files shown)


src/wireviz/DataClasses.py
```diff
@@ -271,6 +271,7 @@ class Cable:
     show_name: Optional[bool] = None
     show_wirecount: bool = True
     show_wirenumbers: Optional[bool] = None
+    show_colorname: bool = True
```
kvid, line 274: I suggest a different name:
```suggestion
    show_wirecolor_text: bool = True
```
Alternatively `show_color_text`, but then someone might expect it to also affect showing the `cable.color` (jacket color) text. Do you want that?

I suggest `text` instead of `name` because the text might be different text representations of the colors, depending on `options.color_mode`.

kvid: This increases the code complexity slightly, but it seems easier to read, and hopefully is easier to understand as well.

I assume you'll also allow the values `True` for showing everything and `False` for showing nothing.

With `None` as default value for `show_colors`, you can set it in `__post_init__()` default `plain` for your single-core cable use case and `True` otherwise.

Issue #404, opened 2024-07-12
[discussion] make position of pin numbers within connectors consistent?


formatc1702:
Split of from https://github.com/wireviz/WireViz/issues/286#issuecomment-2225714107 , image taken from there

![image](https://github.com/user-attachments/assets/822b56af-30a7-471f-92f2-c3c5dbe47b06)

Currently, pin numbers appear on the left or the right of the connector's pin table depending on which side the wires come out.

This was implemented since the very first draft of WireViz and has remained unchanged. As more and more fields can/will be defined on a pin level (color, notes, shorts/bridges, ...), it can cause confusion.

Should pin numbers always appear on the left, to keep the order of the columns consistent across all connectors?
Personally, I am leaning more and more in favor of this solution.

AlanLivingston:
I use these wireviz drawings to document cabling and to assist technicians with construction cables.  Our techs like the pin numbers next to the wires, as that's the information they use most when making the cable.  Many times, they don't care about or refer to the signal name. I'd prefer to move the signal names into the rightmost column after the shorts, so that it's a mirror image of the left connector.

formatc1702:
> Our techs like the pin numbers next to the wires, as that's the information they use most when making the cable.

That was the thinking behind the initial behavior :)

tobiasfalk:
First, I would not talk about left and right but rather see it like an onion, with a core, outside, and inside (currently empty) layers. Currently, the outside layer is always th Nr. and then the core goes lable and jumpers

Second, I think that this outside and inside layers are mirrored if wiers are connected to both sides of the connector.

It would be reasonable to put the lable into the first inside layer.

It would also be good if it would be selectable if these layers are always left, right, both sides or auto(similar to current).

It would also be reasonable if it would be possible to select if the lable is part of the core or layer.

I think that things like the jumpers should remain in the core.

kvid:
@martinrieder wrote:
> I wrote [#404 (comment)](https://github.com/wireviz/WireViz/issues/404#issuecomment-2225983880) under the assumption that `hide_disconnected_pins` is active. IMO this should treat both sides independently to produce a more compact output. Any opinions on this?

- When hiding disconnected pins, it makes sense to hide the whole row because no information about those pins are needed.
[...]

PR #420, opened 2024-09-07
[WIP]Add connector image alignment

martonmiklos:
Added a new 'position' property to connector images which can be used to position images above/right/left of the connector tables:

Above:
kép
Left:
kép
Right:
kép


tobiasfalk:
The left and right option need also to be considered in #404😊

kvid (changes requested):
- Fixing a few bugs (see details in each comment)
- Minor simplifications
- Black formatting (remember running `isort *.py` and `black *.py` before committing your changes)
- Left/right images for `simple` connectors are still not implemented
- Remember that most HTML generation code must later be adapted to #251 changes before a merge-in

kvid (changes requested):
- Fixing a few bugs (see details in each comment)
- Minor simplifications
- Black formatting (remember running `isort *.py` and `black *.py` before committing your changes)
- Left/right images for `simple` connectors are still not implemented (one solution is to only allow `image.position=below` for `simple` connectors)
- Remember that most HTML generation code must later be adapted to #251 changes before a merge-in

PR #440, opened 2025-01-12
Gs/ai generated harnesses

gsubmarine:
ai_harness


formatc1702:
Sorry, I don't see AI integration as a core feature of WireViz. If anything, feel free to write a separate wrapper for it.
Also, what is the point of having the prompt for that single example hardcoded?
Sorry, not adding.

Issue #441, opened 2025-01-15
GitHub Actions fixed by PR #408 are still failing


kvid:
The #408 update of GitHub Actions has been merged into master. I also tried cherry-picking this commit to #416, and the new actions were executed successfully, but two old outdated actions (build 3.7 and build 3.8) seem still to be executed also, and since those fail, it blocks merging. How can I disable old actions after updating old branches with these new actions? @martinrieder @amotl @tobiasfalk @formatc1702 or anyone else?

Could my problem be related to this? https://github.com/orgs/community/discussions/26256

_Originally posted by @kvid in https://github.com/wireviz/WireViz/issues/408#issuecomment-2559735918_

This accepted answer seems relevant: https://stackoverflow.com/questions/67743583/how-can-i-stop-github-actions-from-requiring-old-nodejs-builds

I find the old actions as required before merging into `dev` or `master`, and it seems I can remove them as required, but I don't find the new actions to replace them. What am I missing?

Update: I probably need to cherry-pick the squash commit from mege-in of this PR also into the target `dev` branch before correcting the branch protecting rules, but I must wait testing this until I have access to my PC again.

_Originally posted by @kvid in https://github.com/wireviz/WireViz/issues/408#issuecomment-2559808879_

This issue is currently blocking all PRs from passing required actions.

kvid:
By cherry-picking the commit described above also into the `dev` branch, I finally was able to make the branch protecting rules using the new actions, and it stopped failing due to not finding the old actions anymore.

However, the new actions also fail, like this: https://github.com/wireviz/WireViz/actions/runs/12736147371

"The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04." is the failing action, and the remaining operations were canceled.

It seems the "ubuntu-latest" specifyer started to use ubuntu-24.04 instead of ubuntu-22.04 recently, and ubuntu-24.04 only support Python 3.9, 3.10, 3.11, and 3.12, but has stopped supporting Python 3.7 and 3.8 - and hence those actions are failing. For more details, see https://github.com/actions/runner-images/issues/10636

If we still want actions that uses Python 3.7 and 3.8, then we must specify ubuntu-22.04 for those. Otherwise, we can just drop supporting Python 3.7 and 3.8 from future versions, but then we also need to update README and setup files.

amotl:
[...]

PR #443, opened 2025-01-22
Add SVG Mime Type

TheMDev:
When attempting to add SVG images to connectors and cables, they failed to show up even though the base64 content was correct in the HTML and SVG outputs.
Checking https://www.w3.org/TR/SVG11/intro.html reveals the mime type should be image/svg+xml, not image/svg.
As the code already handles mime subtype replacements for jpeg and tiff, adding svg here made for a simple fix.

### DIFF
--- a/src/wireviz/svgembed.py
+++ b/src/wireviz/svgembed.py
@@ -5,7 +5,7 @@ import re
 from pathlib import Path
 from typing import Union
 
-mime_subtype_replacements = {"jpg": "jpeg", "tif": "tiff"}
+mime_subtype_replacements = {"jpg": "jpeg", "svg": "svg+xml", "tif": "tiff"}
 
 
 # TODO: Share cache and code between data_URI_base64() and embed_svg_images()


kvid:
Thank you for detecting and suggesting a fix for this bug. I tried your fix and I can verify that the HTML output looked correct, but I noticed that the SVG image for a connector is still not shown in the PNG output. I understand that such a problem is outside the scope of this PR, but I wonder if this is also observed by others. JPG, PNG, and GIF images in connectors seems to work fine, but a TIF image was rejected by Graphviz. It would be nice to create a test use case with all image types expected to work, and perhaps a separate issue for unexpected outputs for such a test use case.

The automatic checks are currently failing due to #441, that now should be fixed for PRs with `dev` as target, ~~but I probably need to cherry-pick a commit into `master` to also fix PRs with `master` as target~~.

kvid:
The commit mentioned above is now cherry-picked into `master`, but the automatc checks seem not to be re-executed automatically. Maybe we need to rebase this PR branch on top of the latest `master` (there should not be any conflicts) to re-execute the latest versions of these automatc checks?

@TheMDev - do you want to try doing this yourself, or should I do it?

TheMDev:
I'm sorry for not getting back to you sooner. I cannot replicate your issue where an SVG image does not appear in a PNG output, but I did notice that the PNG output does not follow the same scaling as the HTML and SVG outputs. It seems this is because it uses a different render function that would likely require a rewrite. A more straightforward solution is to convert the SVG output to PNG at export. If you can provide more detailed information on the issue you noticed, I would be more than happy to look into it in this PR, but the scaling issue warrants a separate issue with more discussion on how to resolve it.

Ill rebase my branch onto master here shortly.

kvid:
@TheMDev wrote:
> I'm sorry for not getting back to you sooner.

Nobody are contributing to this project as a payed full-time work and expected to answer quickly.

> I cannot replicate your issue where an SVG image does not appear in a PNG output, but I did notice that the PNG output does not follow the same scaling as the HTML and SVG outputs. It seems this is because it uses a different render function that would likely require a rewrite. A more straightforward solution is to convert the SVG output to PNG at export. If you can provide more detailed information on the issue you noticed, I would be more than happy to look into it in this PR, but the scaling issue warrants a separate issue with more discussion on how to resolve it.

I just created #445 with my preliminary test use case where I also made a comment with my outputs. I don't observe any scaling differences.

Feel free to add a similar comment there with your environment and outputs, and optionally suggest improvements to the test use case.

> 
> I'll rebase my branch onto master here shortly.

Thank's. That helped. All automatic checks have passed now.

kvid (approved):
I'm sorry that I forgot to formally review this PR when testing it several months ago. In my opinion, this is a solid fix to a verified bug, and I find it impossible to find a simpler implementation. 😄

I recommend this as a good candidate for a hot fix release. If there are a couple of other such candidates ready, they might be combined.

PR #469, opened 2025-07-10
Update demo01.yml

ghost:
demo01


kvid (changes requested):
Contributions are very welcome, but please describe what you aim to improve with these suggested changes. Please also read the recommendations in https://github.com/wireviz/WireViz/blob/master/docs/CONTRIBUTING.md

Issue #472, opened 2025-07-11
templates are searched relative to the html file path, not the yaml like the docs claim


ddi-acassidy:
If i run `wireviz -o Output -p lib.yaml 27-*.yaml`, the template I specify in my yaml file is not found. Turns out its looking in the "Output" folder. The docs claim it should be found "In the same folder as the YAML input file."

kvid:
Thank you, @ddi-acassidy for reporting this bug! I have tested this by using `demo02.yml` with a modified custom template name, and can verify that the output folder is searched instead of the input folder (when different).

This bug is not detected automatically because none of the examples/demos/tutorials
- specify a template name from the input folder, and
- specify an output folder different from the input folder.

raquenaengineering:
Hello everyone,
Was this bug fixed?

kvid:
The PR #473 suggests a fix. I reviewed it and verified that the simple cases are resolved, but I suggested a few changes to also resolve the more general cases. You can also try it to see if your use case is resolved.

PR #473, opened 2025-07-11
Fix template search path

ddi-acassidy:
Addresses #472 

the source file path is now forwarded to the Harness object and on to the HTML generator code. Its the last argument with a default to keep API compatibility. Ive also kept the output-relative search location to avoid hurting anyone relying on that behavior

### DIFF
--- a/src/wireviz/Harness.py
+++ b/src/wireviz/Harness.py
@@ -70,6 +70,7 @@ class Harness:
     metadata: Metadata
     options: Options
     tweak: Tweak
+    source_path: Path
 
     def __post_init__(self):
         self.connectors = {}
@@ -697,7 +698,7 @@ class Harness:
             print("CSV output is not yet supported")
         # HTML output
         if "html" in fmt:
-            generate_html_output(filename, bomlist, self.metadata, self.options)
+            generate_html_output(filename, bomlist, self.metadata, self.options, self.source_path)
         # PDF output
         if "pdf" in fmt:
             # TODO: implement PDF output
--- a/src/wireviz/wireviz.py
+++ b/src/wireviz/wireviz.py
@@ -31,6 +31,7 @@ def parse(
     output_dir: Union[str, Path] = None,
     output_name: Union[None, str] = None,
     image_paths: Union[Path, str, List] = [],
+    source_path = None,
 ) -> Any:
     """
     This function takes an input, parses it as a WireViz Harness file,
@@ -115,6 +116,7 @@ def parse(
         metadata=Metadata(**yaml_data.get("metadata", {})),
         options=Options(**yaml_data.get("options", {})),
         tweak=Tweak(**yaml_data.get("tweak", {})),
+        source_path=source_path
     )
     # others
     # store mapping of components to their respective template
--- a/src/wireviz/wv_cli.py
+++ b/src/wireviz/wv_cli.py
@@ -144,6 +144,7 @@ def wireviz(file, format, prepend, output_dir, output_name, version):
             output_dir=_output_dir,
             output_name=_output_name,
             image_paths=list(image_paths),
+            source_path=file
         )
 
     print()
--- a/src/wireviz/wv_html.py
+++ b/src/wireviz/wv_html.py
@@ -21,14 +21,19 @@ def generate_html_output(
     bom_list: List[List[str]],
     metadata: Metadata,
     options: Options,
+    source: Union[str, Path] = None,
 ):
     # load HTML template
     templatename = metadata.get("template", {}).get("name")
+    template_search_paths = [ Path(filename).parent, Path(__file__).parent / "templates"]
+
+    if source is not None:
+        template_search_paths.insert(0, Path(source).parent)
+
     if templatename:
         # if relative path to template was provided, check directory of YAML file first, fall back to built-in template directory
         templatefile = smart_file_resolve(
-            f"{templatename}.html",
-            [Path(filename).parent, Path(__file__).parent / "templates"],
+            f"{templatename}.html", template_search_paths
         )
     else:
         # fall back to built-in simple template if no template was provided


kvid:
Thank you @ddi-acassidy for this suggested fix for #472. However, I think we need to discuss the implementation a bit.
1. It seems to fix the issue for simple cases where the template name is specified in the main YAML input file, but not if the template name is specified in a prepend YAML file located in a different folder than the main YAML input file.
2. The problem described above is very similar to searching for images with a relative path, and I think we can reuse the `image_paths` list that was created to resolve the same issue for images with a relative path. If this can be verified, then I suggest renaming it to `input_paths` and use it for all relative paths (to images, templates, and any other separate input files specified by a relative path).
3. If my suggestion is accepted and verified, then PR #444 also needs to be adjusted accordingly.
4. Please argue against my suggestion, and search for a use case where my suggestion will not be a good solution.
5. Please remember running `isort` and `black` on Python files before committing your changes - as described together with other recommendations in [CONTRIBUTING.md](https://github.com/wireviz/WireViz/blob/master/docs/CONTRIBUTING.md).

ddi-acassidy:
I hadnt thought of that case. If template is specified in the include file then the search path needs to be different from when its in the main file. Fixing that requires parsing the include file separately instead of just concatenating it with the other file.

kvid:
> I hadnt thought of that case. If template is specified in the include file then the search path needs to be different from when its in the main file. Fixing that requires parsing the include file separately instead of just concatenating it with the other file.

I understand your reasoning, and the same argument was used when discussing solutions  to the similar image relative path issue. However, a pragmatic approach was finally selected that adds folders of all input YAML files into a list of paths, by searching relative image paths relative to all folders in this list, and use the first one found. I suggest using the same approach when searching templates for consistency.

kvid:
> I hadnt thought of that case. If template is specified in the include file then the search path needs to be different from when its in the main file. Fixing that requires parsing the include file separately instead of just concatenating it with the other file.

I understand your reasoning, and the same argument was used when discussing solutions  to the similar image relative path issue. However, a pragmatic approach was finally selected that adds folders of all input YAML files into a list of paths, by searching relative image paths relative to all folders in this list, and use the first one found. I suggest using the same approach when searching templates for consistency.

ddi-acassidy:
Got it, I wasnt aware of how those two bullet points fit together to solve the problem. I'll look into updating my PR with those changes when I have time

Relatedly, I put together support for Jinja templates as an alternative to the ersatz html comment template system available at the moment. It allows much more flexibility with HTML templates, like variable sized revision blocks and arbitrarily structured metadata, any interest in me cleaning that up and PRing it?

kvid:
> Got it, I wasnt aware of how those two bullet points fit together to solve the problem. I'll look into updating my PR with those changes when I have time

Please verify that reusing the existing `image_paths` variable will resolve the issue before (in a separate commit) renaming the variable to a generic name.

> 
> Relatedly, I put together support for Jinja templates as an alternative to the ersatz html comment template system available at the moment. It allows much more flexibility with HTML templates, like variable sized revision blocks and arbitrarily structured metadata, any interest in me cleaning that up and PRing it?

See also the existing PR #382 for Jinja.

src/wireviz/Harness.py
```diff
@@ -70,6 +70,7 @@ class Harness:
     metadata: Metadata
     options: Options
     tweak: Tweak
+    source_path: Path
```
kvid, line 73:
```suggestion
    input_folders: Union[str, Path, List]
```

src/wireviz/Harness.py
```diff
@@ -697,7 +698,7 @@ def output(
             print("CSV output is not yet supported")
         # HTML output
         if "html" in fmt:
-            generate_html_output(filename, bomlist, self.metadata, self.options)
+            generate_html_output(filename, bomlist, self.metadata, self.options, self.source_path)
```
kvid, line 701:
```suggestion
            generate_html_output(
                filename, bomlist, self.metadata, self.options, self.input_folders
            )
```
The splitting into multiple lines is due to `black` formatting.

src/wireviz/wireviz.py
```diff
@@ -31,6 +31,7 @@ def parse(
     output_dir: Union[str, Path] = None,
     output_name: Union[None, str] = None,
     image_paths: Union[Path, str, List] = [],
+    source_path = None,
```
kvid, line 34:
```suggestion
```

src/wireviz/wireviz.py
```diff
@@ -115,6 +116,7 @@ def parse(
         metadata=Metadata(**yaml_data.get("metadata", {})),
         options=Options(**yaml_data.get("options", {})),
         tweak=Tweak(**yaml_data.get("tweak", {})),
+        source_path=source_path
```
kvid, line 119: I suggest reusing the existing `image_paths` that already has collected all input folders:
```suggestion
        input_folders=image_paths,
```

src/wireviz/wv_cli.py
```diff
@@ -144,6 +144,7 @@ def wireviz(file, format, prepend, output_dir, output_name, version):
             output_dir=_output_dir,
             output_name=_output_name,
             image_paths=list(image_paths),
+            source_path=file
```
kvid, line 147:
```suggestion
```

src/wireviz/wv_html.py
```diff
@@ -21,14 +21,19 @@ def generate_html_output(
     bom_list: List[List[str]],
     metadata: Metadata,
     options: Options,
+    source: Union[str, Path] = None,
```
kvid, line 24:
```suggestion
    input_folders: Union[str, Path, List] = [],
```

src/wireviz/wv_html.py
```diff
@@ -21,14 +21,19 @@ def generate_html_output(
...
 ):
     # load HTML template
     templatename = metadata.get("template", {}).get("name")
+    template_search_paths = [ Path(filename).parent, Path(__file__).parent / "templates"]
+
+    if source is not None:
+        template_search_paths.insert(0, Path(source).parent)
+
```
kvid, line 32:
```suggestion
```

src/wireviz/wv_html.py
```diff
@@ -21,14 +21,19 @@ def generate_html_output(
...
     if templatename:
         # if relative path to template was provided, check directory of YAML file first, fall back to built-in template directory
         templatefile = smart_file_resolve(
-            f"{templatename}.html",
-            [Path(filename).parent, Path(__file__).parent / "templates"],
+            f"{templatename}.html", template_search_paths
```
kvid, line 36: Remember inserting `from wireviz.wv_bom import make_list` in the import section above and run `isort *.py` to format and sort them. (Github don't allow me to directly suggest changes at lines more than a few lines away from existing PR changes.)
```suggestion
            f"{templatename}.html",
            make_list(input_folders) + [Path(__file__).parent / "templates"],
```
My suggestion here does not add the output folder in the list when different because I don't see any use case where that might be useful, but please argue against my suggestion.

kvid (changes requested):
I have in this review detailed my suggestions to reuse the existing `image_paths` list that already has collected all input folders to resolve the issue with more than one input YAML file at different folders. If you agree to this (after testing several use cases), then as a separate commit, renaming `image_paths` to `input_folders` might be a good change to reflect the more generic usage.

tbornon-sts:
Hi,
I'm interested by that change. Is there anything holding it back from being merged ?

kvid:
@tbornon-sts wrote:
> Hi, I'm interested by that change. Is there anything holding it back from being merged ?

As you can see above, I have reviewed this suggested change and verified that the simple cases are resolved, but I suggested a few more changes to also resolve the more general cases.

Feel free to download/install the suggested changes from [this branch](https://github.com/ddi-acassidy/WireViz/tree/path-fix) yourself to test if your use case is resolved.

If the author doesn't find time to do the work, then we probably should consider merging it as is and add the additional changes on top.

tbornon-sts:
Sorry, it took me quite some times to review your changes. However, I've just tried it and it doesn't fit my issue nor fit with my workflow.

The proposed fix is great if your template is always stored at the same level relatively to your YAML file. However, for me it's not the case. The template is at the root of my project while the yaml files are stored in different levels of folders. 

Tree of my project :

```
.
├── src
│   ├── Folder1
│    |      ├── assets
│    |       |      ├── A.png
│    |      ├── A.yaml
│    |      ├── B.yaml
│   ├── Folder1
│    |      ├── C.yaml
│    |      ├── D.yaml
│   ├── E.yaml
│   ├── F.yaml
├── build
└── library
    ├── accessories.yaml   <-- File included with --prepend flag
    ├── cables.yaml           <-- File included with --prepend flag
    ├── connectors.yaml   <-- File included with --prepend flag
    ├── contacts.yaml       <-- File included with --prepend flag
    └── template
        └── template.html   <-- Template file I want to include in all my YAML
```

Regarding the issue, I see two possibilities to improve it :
1. As you mentionned, if a file is prepended, then add it to the folders where we search
2. Add the flag I proposed in #444

Issue #491, opened 2026-01-04
[feature] Revision place holder in html for latest revision


ishaid:
Hello,

I’ll preface this by saying that I’ve already implemented the change and am preparing a PR. I’m opening this issue as well, per the contribution guidelines.

Problem / missing feature

As far as I can tell, if you want to insert a revision identifier into an HTML file, you currently have to specify the revision explicitly in the placeholder. This is problematic for two reasons:
	•	You need to update the HTML template every time the revision changes.
	•	If multiple files use the same template but are on different revisions, there’s no good way to display the current revision for each harness in its drawing.

I’ve attached an image showing this in my title block at the end of the post.

Proposed solution

I’ll add more detail in the PR, but the gist is to introduce a placeholder like:
`

`
and populate it with the latest revision from the revisions section of the .yml.

<img width="598" height="256" alt="Image" src="https://github.com/user-attachments/assets/b1c2663f-1ca3-4bae-a0d9-c07d443cd12e" />

tbornon-sts:
Hi,

The standard template already supports revisions : https://github.com/wireviz/WireViz/blob/master/src/wireviz/templates/din-6771.html#L265
You then just have to list them in your yaml

For example :

```yaml
metadata:
    ...
    revisions:
        A: 
            changelog: Initial release
            date: 23/05/2025
        B:
            changelog: Update PNs for fab
            date: 22/01/2026
    ...
```

ishai-l5:
@tbornon-sts Hey, yes maybe I wasn't clear, revision do already exit but there is no placeholder that gets populated with the latest revision. This is what the feature is talking about.

It's a common industry practice to have the latest revision denoted in addition to the revision table and without a placeholder for the latest revision this isn't possible.

PR #492, opened 2026-01-04
Added <!-- %revision% --> place-holder

ishaid:
Hello,

This addresses issue [#491](https://github.com/wireviz/WireViz/issues/491) (which I created per the contribution guidelines).

I've added a `<!-- %revision% -->` place holder which gets populated by the latest revision using the following logic:

`
def _get_latest_revision(metadata: Dict) -> str:
    if not "revisions" in metadata:
        return ""
    revision = list(metadata.get("revisions"))[-1]
    return revision
`

In addition to what I mentioned in the issue this also allows adding a revision indicator of the current revision without needing the entire table.

I've added the place holder to the relevant README.md.


kvid:
@ishaid - Thank you for your contributions! With your latest commits, it seems you have now added 3 independent suggestions (revision placeholder, pyprojects.toml and new docs) in this PR. I therefore suggest you split out the latest commits into two new PRs to enable independent reviews and discussions.

If this is hard because one PR then will depend fully on the other, it is possible to include all the dependent commits in both, and make a note (in the PR description) of which (probably leading) commits are cherry-picked from another PR (which one) due to dependencies, and therefore not directly part of the PR.

Issue #465, opened 2025-05-23
[Bug] Loops on connectors number vs list index issue


ryane67:
When using loops on connectors, the check to draw ensures the connector number is in the list, however when they are drawn, the connection between the points is made from what appears to be the list index and not by the pin number.

```
connectors:
  AP-DISCO:
    pins: [1]
    pinlabels: [GND_SW]
  APRS1-J281:
    pins: [0,1,2,3,4,5,8,9,10,11,12,13,14,15]
    pinlabels: [SHELL, CANHI, CANLO,CANTERM1, CANTERM2,IDSTRAP_1,IDSTRAP_4,PWRGND, ACFTPWR,TRIMIN1, TRIMIN2, TRIMOUT1, TRIMOUT2, APDISCO]
    <<: [*db15_m, *can_device]
    loops:
      - [4,5]
      - [5,8]
```

![Image](https://github.com/user-attachments/assets/13da514e-407c-4dc0-83a2-76b5ad3d907d)

The intent here would be to realistically use 3,4 and 5,8 as the loops however I used 4,5 when testing whether it would use the index of pins in the list to create the graph.

kvid:
Thank you for reporting this unfortunate bug with the current `loops` implementation. It's heavily related to #432 that is already marked with a [TODO in the code](https://github.com/wireviz/WireViz/blob/v0.4.1/src/wireviz/DataClasses.py#L207).

Issue #432, opened 2024-11-29
Loops specification can't use pin labels?


lutorm:
I have a wire that should connect two pins on the same connector, and I discovered the "loops" feature. However, when I attempted to use it, I got this error:

> exception: Unknown loop pin "6" for connector "P4601"!

From some experimentation, it appears that the loops lists only understands pin numbers, not pin labels? Is this intentional? It seems to go against the pattern used everywhere else where you can use numbers and labels interchangeably.

(This feature also seems a bit half-baked, honestly. I would need to specify the color and gauge of the wire used, so I'd need to make up a cable section for it anyway so these can be specified.)

kvid:
I agree the current `loops` feature has limitations. Possible improvements are discussed in several issues, e.g. #350 and #439.

kvid:
It seems the generic part of [the pin mapping code](https://github.com/wireviz/WireViz/blob/v0.4.1/src/wireviz/Harness.py#L112-L129) in `Harness.connect()` needs to be moved into a new `Connector.pin_index()` method to be available from both `Harness.connect()` and from `Connector.__post_init__()` (the latter is where loop pins are processed).

Using the same pin mapping for both connections and loops should resolve both this issue and #465.

To enhance code readability, we should also clearly state where we use pin indexes (always zero based?) and where we use pin names (from `pins` attribute) and any other valid pin identifier (e.g. `pinlabels`). We can use different type hints for this, but I hope we also can use variable names that also helps the reader to understand.

Issue #328, opened 2023-12-29
[bug] Unconnected components are ignored by v0.4-dev


kvid:
The current `dev` branch (after merging in multiple PRs this summer) is no longer able to document unconnected connectors or cables - as has been possible with all releases (at least v0.2 and later).

This is a minimal example to document a single connector without any connections:
```yaml
connectors:
  X1:
    pinlabels: [GND, VCC]
```

With the current `dev` branch, this results in an empty diagram - which is not very useful!

Why is the algorithm that generates the diagram rewritten to only add components that are referred to from connections? That makes it impossible to document e.g. a loop-back connector, which is not an unusual use case. See also the slightly related #217.

The PR #251 on top of `dev` does not seem to help with this issue.

kvid:
If this new fuctionality is intended as a feature, then it might be a solution to #294. However, in that case, it should not be the default option, IMHO, because I find it counterintuitive unless consiosly activating it as an optional feature.

formatc1702:
Thanks for bringing this up.

The easy fix with the current code, is to define a connection set, with only that connector included:
```yaml
connectors:
  X1:
    pinlabels: [GND, VCC]

connections:
  - - X1
```

The question is where `harness.add_connector()` is called, which currently only happens inside `for connection_set in connection_sets:` after resolving designators etc., instead of during initial YAML parsing.

I am OK with restoring the old behavior, feel free to submit a PR and adding the option as you suggest. Bear in mind the same would apply to unconnected `cables` as well.

sb424dat:
I you want I can add this to the documentation and try some test cases on my machine!

formatc1702:
I am coming back to this issue, and I can now offer a better explanation.
TL;DR: it's a feature, not a bug; but I agree it was an unexpected side effect of that feature.

Previously, components were immediately generated unless they were explicitly meant for autogeneration, then added to the harness, and thus rendered: https://github.com/wireviz/WireViz/blob/b0d0070f08247e132b11bf45e617da9e8e1881f1/src/wireviz/wireviz.py#L64-L68

[...]

PR #496, opened 2026-04-24
Fix loopback rendering (issue #439)

### LINKED ISSUE #439: Add a way to add a loopback wire, add example of it
I want to add a tiny loopback wire as shown in the following diagram (pink jagged line).

The rest of the diagram is below. 

This would be great to add as an example if it's currently possible; otherwise, it'd be a great feature to add! 

image

<details>

```yaml
connectors:
  CN1 (AC Input):
    type: JST B3P-VH on PCB, JST SVH-21T-P1.1 on wire, JST VHR housing
    subtype: female
    pinlabels: [Neutral, N/C, Line]
  CN2 (DC Output):
    type: JST B8P-VH on PCB, JST SVH-21T-P1.1 on wire, JST VHR housing
    subtype: female
    pinlabels: [+V, +V, +V, +V, -V, -V, -V, -V]
  CN3 (Power Good):
    type: JST B4B-XH on PCB, JST SXH-001T-P0.6 on wire, JST XHP housing
    subtype: female
    pinlabels: [PG, GND, Neg Sense, Pos Sense]
  CN901 (Standby):
    type: JST B-XH on PCB, JST SXH-001T on wire, JST XHP housing
    subtype: female
    pinlabels: [N_PS_ON, GND, 5VSB, GND]

  AC Input:
    subtype: female
    pinlabels: [Line, N/C, Neutral]

  Main Connector:
    subtype: female
    pinlabels: [V1, V1, V1, RTN, RTN, RTN, Neg Sense, Pos Sense, Remote PS ON]

cables:
  AC Wire:
    # gauge: 0.25 mm2
    # length: 0.2
    color_code: DIN
    wirecount: 2

  Main Wire:
    # gauge: 0.25 mm2
    # length: 0.2
    color_code: DIN
    wirecount: 9

  Pigtail Loopback:
    # gauge: 0.25 mm2
    # length: 0.2
    color_code: DIN
    wirecount: 1

connections:
  -
    - CN1 (AC Input): [3, 1]
    - AC Wire: [1, 2]
    - AC Input: [1, 3]

  -
    - CN2 (DC Output): [1-3, 5-7]
    - Main Wire: [1-6]
    - Main Connector: [1-6]

  -
    - CN3 (Power Good): [3, 4]
    - Main Wire: [7, 8]
    - Main Connector: [7, 8]

  -
    - CN901 (Standby): [1]
    - Main Wire: [9]
    - Main Connector: [9]
```
</details>

eisbaw:
Closes #439.

Three root causes fixed in the loop-handling path:

1. **Loop-only connectors silently dropped.** A connector declared with `loops:` but not referenced in any connection set was treated as an "unused template" and never instantiated. Such connectors are now auto-instantiated with an `Info:` message noting the promotion.

2. **All loops forced onto a single side.** The previous `create_graph` picked one side for the whole connector from `ports_left`/`ports_right` and emitted all loop edges there. On two-sided connectors this produced stray dangling arcs on the wrong side. The dead-end `raise Exception("No side for loops")` could also trigger on connectors whose only activations came from loops.

   Now each pin records its activated sides in a new `Connector.pin_sides: Dict[Pin, Set[Side]]`, and `Connector.resolve_loops()` picks per-loop endpoints by intersecting the two pins' known sides. Disjoint sides produce a cross-connector loop; loop-only pin pairs fall back to an existing port column; the no-information default is RIGHT.

3. **Loop edge ports used pin numbers instead of positions.** Pin ports in the connector node are emitted as `p{pinindex+1}{l|r}` (1-based positions). Cable- and mate-edge emission correctly translate pin number to position via `pins.index(pin) + 1`. Loop edge emission did not -- it built port IDs directly from the pin number in `loops: [[a, b]]`. When pin numbers happened to equal their positions (default `pincount=4` yielding `[1,2,3,4]`) the two agreed by accident. On connectors with an explicit non-sequential `pins:` list (e.g. a D-Sub wired for differential pairs), the loop edges referenced nonexistent port IDs and GraphViz silently drew them at the node centre, producing weird arcs routed over the top of the connector. This bug was pre-existing but only surfaced once loop rendering was exercised more broadly.

## Files touched

- `src/wireviz/DataClasses.py` -- `Connector.pin_sides`, `resolve_loops()`
- `src/wireviz/Harness.py` -- call `resolve_loops()` before node emission, per-loop `(side_a, side_b)` edges, pin-position lookup on the loop endpoints
- `src/wireviz/wireviz.py` -- auto-instantiate loop-only connectors with an Info log
- `examples/loopback_bug.yml` -- driving example: loop-only connector, loops on a two-sided connector, loops on a single-side connector
- `examples/loopback_return.yml` -- three-connector return-path illustration; documents that cross-connector "loopbacks" are modelled as cables, not `loops:`
- `examples/loopback_nonseq.yml` -- regression witness for the pin-number-vs-pin-position bug

## Test plan

- [x] All 14 `examples/ex*.yml` and 8 `tutorial/tutorial*.yml` still render without errors or warnings.
- [x] New `examples/loopback_bug.yml` renders with X1 (loop-only) instantiated, X2's loop on its right side next to the cable it shares pins with, X3's loop on its right side.
- [x] New `examples/loopback_nonseq.yml` renders with loop edges referencing pin-table positions 1-2 and 3-4 (for pin numbers [1, 14] and [3, 16]), visible as coils between adjacent pin rows.
- [ ] Reviewer to verify BOM/HTML outputs are unaffected (this change touches only graph emission for the loop edges and auto-instantiation).

## Known follow-ups (not in this PR)

- `style: simple` + `loops:` emits `port pXr unrecognized` warnings in GraphViz because simple-style nodes don't render a pin table; this is a pre-existing issue and would be cleanest to reject at `Connector.__post_init__`.
- `loops:` referencing `pinlabels:` instead of pin numbers is still rejected by post-init with a clear error (TODO in the pre-existing code).
- `ports_left`/`ports_right` are now strictly derivable from `pin_sides` and could be made `@property` in a follow-up; left as-is to keep this change minimally invasive.


kvid:
@eisbaw - Thank you very much for this contribution. I look forward to test this, but have not had the time yet.

About your root cause number one (Loop-only connectors silently dropped) - I suspect this is as intended. Please read #328 where I reported something similar as a bug, then Daniel (the owner) argued why it is intended, and convinced me that it is a good design choice. The autogeneration advantages are greater than the minor disadvantage when needing an unconnected connector. He has described it in the [syntax documentation](https://github.com/wireviz/WireViz/blob/master/docs/syntax.md#unconnected-components), and to inform any unaware users, all unconnected entries are listed in a run-time warning: `"Warning: The following components are not referenced in any connection set:"`

kvid:
Please don't regard this as a hostile attack on your suggested changes. I just want to understand your use case and what is a better functionality for that.

@eisbaw wrote:
> Closes #439.

Are you sure this is the correct issue number? #439 is about wires between two different connectors, but your PR seems only to suggest changes about loops between pins on the same connector.

However, the reported issues #432 and #465 seem related to your root cause 3.

> 1. **Loop-only connectors silently dropped.**

See my previous comment above and #328.

> 2. **All loops forced onto a single side.** The previous `create_graph` picked one side for the whole connector from `ports_left`/`ports_right` and emitted all loop edges there.

I understand you feel this is wrong. Please explain why. What reasons in your use case drive you to prefer a loop from pin X to be on the same side as another connection from the same pin?

Remember that connections on both sides are just a drawing technique to reduce crossing lines in the diagram - not representing a physical two-sided connector

> On two-sided connectors this produced stray dangling arcs on the wrong side.

Please define "wrong side" in your use case.

> The dead-end `raise Exception("No side for loops")` could also trigger on connectors whose only activations came from loops.

Please provide an example YAML input that triggers this exception with the currently latest release.

>    Now each pin records its activated sides in a new `Connector.pin_sides: Dict[Pin, Set[Side]]`, and `Connector.resolve_loops()` picks per-loop endpoints by intersecting the two pins' known sides. Disjoint sides produce a cross-connector loop;

Please describe what you mean by a "cross-connector loop" in this case? Do you suggest a spline wire between a left pin and a right pin on the same connector?

> loop-only pin pairs fall back to an existing port column; the no-information default is RIGHT.

AFAIK, the existing default is the LEFT side. Is RIGHT a better default in your use case?

PR #509, opened 2026-08-18
feat(options): expose ranksep, nodesep, and wire_thickness

### LINKED ISSUE #7: Can units be in inches?
Is there a way to set unit to inches?

mkhnsn:
Closes #7

Exposes three previously hardcoded layout-density knobs in `options`:

- `ranksep` (default 2) — inches between ranks; setting ~1 visibly narrows wide LR drawings
- `nodesep` (default 0.33) — inches between nodes within a rank
- `wire_thickness` (default 2) — scales edge `penwidth` and the in-node colour-stripe rows together, so wires stay flush where they meet cable tables; helps pale wires survive print scale

Compatibility: defaults produce **byte-identical** GV source (verified by diffing regenerated `demo02.gv` against the committed one, and asserted in `tests/test_layout_options.py`). `penwidth` is only emitted when thickness differs from the default, since `style=bold` already implies 2pt.

Validation rejects non-numeric and non-positive values.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WzBaE7rN4STA7Juixad83t


mkhnsn:
Apologies — opened against upstream by mistake; this belongs on our fork.

